PR #2, opened 2024-06-05
Functions from corridor edge delineation

cforgaci:
I translated the functions from the corridor delineation prototyping notebook into functions. I also moved the narrative from the notebook to a package vignette. This is still rough and quite broad, so I suggest we focus on overall functionality in this PR, merge soon, and continue with refining and refactoring in more granular branches.


fnattino (changes requested):
Great work, looks already in great shape! Nice structure, maybe when the package will grow we might thing in defining a geometry  and a network modules, to distribute some of the function currently in utils and preprocess - but for now it makes a lot of sense as it is. I could run the vignette smoothly using the environment provided, so all seems to be consisntent.

We should maybe remove the `hello.R` initial file?

cforgaci:
Thanks for the review @fnattino. 

> We should maybe remove the `hello.R` initial file?

I removed it now.

Let me know when we are ready to merge this branch.

fnattino (approved):
Looks very ready to me :)

PR #4, opened 2024-06-25
Add pkgcheck to repository

cforgaci:
I also added `pkgcheck` which needs to be passed if we want to submit the package for review on rOpenSci, as indicated in the proposal.


cforgaci:
> I guess these could also be addressed in new PRs.

Yes, let's address these in separate PRs as they are not formally part of pkgcheck.

fnattino (commented):
The error it raises is linked to the fact that the package only contains a dummy file and no public functions, correct? In that case I think this can also be merged!

cforgaci:
> The error it raises is linked to the fact that the package only contains a dummy file and no public functions, correct? In that case I think this can also be merged!

I am not sure, but in any case I would explore this error in the context of the (now) more complete main branch. I will merge.

Issue #13, opened 2024-07-01
City boundaries


cforgaci:
The key:value pair "place:city" only works for some cases, our first example Bucharest included. In other cases "boundary:administrative" works instead. We need to generalize this.

fnattino:
Maybe worth to explore: GADM dataset for "harmonised " administrative boundaries https://gadm.org/index.html

cforgaci:
> Maybe worth to explore: GADM dataset for "harmonised " administrative boundaries https://gadm.org/index.html

I had a look at this dataset with a downloaded version of all admin boundaries (1.6GB).

This is still not perfect. `NAME_1` works for Bucharest, while for Paris and Berlin the `NAME_2` variable has the name of the city. However, we can search across different levels and return a geometry when it is found. This works for the examples we've seen so far.

```r
gadm_410[gadm_410$COUNTRY == "Romania" & gadm_410$NAME_1 == "Bucharest",]
gadm_410[gadm_410$COUNTRY == "France" & gadm_410$NAME_2 == "Paris",]
gadm_410[gadm_410$COUNTRY == "Germany" & gadm_410$NAME_2 == "Berlin",]
```

cforgaci:
This issue is partially solved in #15. We still need to find a more robust solution.

cforgaci:
I am closing this issue for now. If we encounter issues with the new function we can reopen.

PR #15, opened 2024-08-26
Add function to get OSM city boundary

cforgaci:
Looking at other cities, we see that city boundaries are not always stored under the same OSM tag. This function searches for a known set of key:value pairs and to get the city boundary.

This still uses OSM, as opposed to the GADM dataset alternative desceribed in #13. Maybe it is better to stick to OSM data throughout the package?


cforgaci:
The function `get_osm_city_boundary()` is quite wordy and can be improved but it does return a city boundary for Bucharest, Paris, Berlin and some other cases I tested as well.

R/retreive-osm.R
```diff
@@ -43,11 +29,54 @@ osmdata_as_sf <- function(key, value, bb) {
...
+#'
+#' @return An sf object with the city boundary
+#' @export
+#'
+#' @examples
+#' get_osm_city_boundary("Bucharest")
+get_osm_city_boundary <- function(city_name) {
+  city_boundary <- tryCatch(
+    {
+      CRiSp::get_osmdata(city_name, "place", "city")$osm_multipolygons$geometry
```
fnattino, line 49: After trying to run this for Berlin:
```
Berlin <- CRiSp::get_osm_city_boundary("Berlin")
```
I got two features as output, with the two geometries representing administrative boundaries for Berlin and Postdam:
```
# Berlin
Geometry set for 2 features 
Geometry type: MULTIPOLYGON
Dimension:     XY
Bounding box:  xmin: 12.88678 ymin: 52.33824 xmax: 13.76116 ymax: 52.67551
Geodetic CRS:  WGS 84
MULTIPOLYGON (((13.09076 52.41156, 13.09144 52....
MULTIPOLYGON (((13.15028 52.56771, 13.15014 52....
```
plot

Filtering the dataframe here seems solve the issue for this specific case, but I am not sure how robust this is (and we are somehow enforcing the use of English names for the cities)?

```suggestion
      CRiSp::get_osmdata(city_name, "place", "city")$osm_multipolygons |> dplyr::filter(.data$`name:en` == city_name) |> st_geometry()
```

fnattino (commented):
Thanks @cforgaci, providing utility functions to extract administrative boundaries only from OSM sounds good (we can still offer the possibility to inject boundaries from other sources in the functions that make use of the boundaries..).

Just one comment below, I don't know how much of an edge case this is.

tests/testthat/test-retreive-osm.R
```diff
@@ -0,0 +1,21 @@
+test_that("City boundary is correctly retreived", {
+  bucharest_boundary_path <- testthat::test_path("testdata",
```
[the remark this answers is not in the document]
cforgaci: Yes, definitely.

R/retreive-osm.R
cforgaci: This is an important edge case, I am sure we will come across this again. Further filtering can be made inside the function. I'll try to update the function to return the correct boundary for Berlin as well.

cforgaci:
Thanks, @fnattino!

> providing utility functions to extract administrative boundaries only from OSM sounds good (we can still offer the possibility to inject boundaries from other sources in the functions that make use of the boundaries..).

👍

> Just one comment below, I don't know how much of an edge case this is.

I'll try to solve that edge case before merging.

R/retreive-osm.R
cforgaci: I updated the function to work for the Berlin case. I made the function more compact and a bit more robust; it can be further improved, but that can be done in a separate PR. I will merge this one.

PR #16, opened 2024-08-26
Clean assignment

fnattino:
Not sure why this is assignment is here - to be removed?


fnattino:
Oh, now I see the reason for the assignment - closing this.

PR #18, opened 2024-08-27
Setting up testing for the repository

### LINKED ISSUE #12: Package data for examples and tests
Should we add a simple dataset to the package that we can use in examples and tests?

fnattino:
I think the repository should include enough data not to make queries to the Overpass API while running continuous integration. We should probably be storing the data in a form that is as close as possible to the way in which raw (OSM) data is provided, in order to test functionality of the cleaning steps. We should also consider storing cleaned data later on (e.g. cleaned street network), in order to test the functionalities of the steps that follows.    

- [x] add test data to the repository for realistic testing (considering Bucharest, our study case for the moment).
- [x] include the script that downloads the data, in order to be able to update the test datasets in future.
- [x] tests running Overpass API queries should be skipped ~unless they are specifically flagged to run~ (EDIT: there is a nice `testthat` functionality which suppresses tests on CI: `skip_on_ci` ) - include example of a test that needs to be skipped.
- [x] all other tests should instead run using the testdata - include an example of this too.   

Closes #12.


R/preprocess.R
```diff
@@ -28,4 +28,6 @@ create_network <- function(data, crs = NULL) {
   if (!is.null(crs)) {
     net <- net |> sf::st_transform(crs)
   }
+
+  net
```
fnattino, line 32: This function was not returning the output - a small fix outside the scope of the PR to enable tests on this function to run

tests/testthat/testdata/download_testdata.R
```diff
@@ -0,0 +1,85 @@
...
+  waterbody <- get_waterbody(bbox, waterway)
+
+  save_sf(highways, filename = sprintf("highways_%s.gpkg", city))
+  save_sf(railways, filename = sprintf("railways_%s.gpkg", city))
+  save_sf(city_boundary, filename = sprintf("city_boundary_%s.gpkg", city))
+  save_sf(waterway, filename = sprintf("waterway_%s.gpkg", city))
+  save_sf(waterbody, filename = sprintf("waterbody_%s.gpkg", city))
+}
+
+download_testdata()
```
cforgaci, line 85: creating and calling `download_testdata()` is not necessary here as the function is only used once

tests/testthat/testdata/download_testdata.R
```diff
@@ -0,0 +1,85 @@
...
+  highways
+}
+
+get_railways <- function(bbox) {
+  railways <- CRiSp::osmdata_as_sf("railway", "rail", bbox)
+  highways <- railways$osm_lines |>
+    sf::st_geometry() |>
+    sf::st_transform(epsg_code)
+
+  highways
```
cforgaci, line 30: The function will return the value of the last evaluated expression, even if it was assigned to an object, so this is not necessary to explicitly call `highways`. This applies to the other functions as well.

At least this is what I've been doing so far, I hope I am not mistaken.

R/preprocess.R
cforgaci: I do not think this is necessart though. Did you have issues with this? See my comment below.

tests/testthat/testdata/download_testdata.R
cforgaci: See this basic example for the same behaviour:

```r
my_function <- function(x) {
  y <- x + 1  # y is created with assignment
}

result <- my_function(5)
result  # Output will be 6
```

cforgaci (commented):
Looks good, thanks! I left a couple of comments on the use of functions.

tests/testthat/testdata/download_testdata.R
fnattino: Thanks for the clarification.  You are definitely right! Using your example, I was fooled by the fact that just calling `my_function(5)` does not print any output (but it is just the printing that is missing, assigning the function output to a variable shows that the value is properly returned).

R/preprocess.R
fnattino: Uhm, however here the function does not return the expected output unless the crs (optional) argument is specified. Maybe the behaviour is different with the if-block closing the function?

See following example:

```r
f <- function(x, opt = FALSE) {
  y <- x + 1

  if (opt) {
    y <- y + 1
  }
} 

a <- f(1)  # a is NULL
a <- f(1, opt = TRUE)  # a is 3 
```

cforgaci: Indeed. It is different if the body of the function ends with an if block. Then let's keep it as you suggest. Thanks.

cforgaci (approved):

PR #24, opened 2024-08-31
Package data

cforgaci:
I initialised the package data with the city boundary and a couple of other layers of Bucharest. The data is stored as a list with a named element for each layer we want to store (e.g., `bucharest$boundary` for the layer I added). We can add other layers to that list via the data generation script in `data_raw/bucharest.R` which will update the package data in `data/bucharest.rda` when it is run.

If the list becomes too big with all the layers, we can replace the list with separate `rda` files.


fnattino:
Nice way to serialise data and include it as part of the package! Since this is actually more or less the same data as in testing, maybe this can also replace the testdata as well?

cforgaci:
> Since this is actually more or less the same data as in testing, maybe this can also replace the testdata as well?
Yes, I think so, as the data is used in the vignettes, on the user side, as well.

PR #25, opened 2024-09-10
Function to get UTM zone for a given sf object

cforgaci:
This utility function returns the UTM zone of a given sf object. The longitude of the centroid is used to get the UTM zone.


R/utils.R
```diff
@@ -34,3 +34,23 @@ calc_weights <- function(net) {
...
+get_utm_zone_epsg <- function(x) {
+  if (!"sf" %in% class(x)) {
+    stop("x must be an sf object")
+  }
+
+  long <- x |>
+    sf::st_transform(4326) |>
+    sf::st_union() |>
+    sf::st_centroid() |>
+    sf::st_coordinates()
```
fnattino, line 53: Probably it does not matter too much, but if `x` contains many/complicated geometries, the reprojection/union/calculation of the centroid can become expensive. What about the following alternative approach, not exactly equivalent, but maybe good enough for the purpose?

Also I suggest to rename the variable to `coords`, see following comment.

```suggestion
  coords <- x |>
    sf::st_bbox() |>
    sf::st_as_sfc() |>
    sf::st_centroid() |>
    sf::st_transform(4326) |>
    sf::st_coordinates()
```

R/utils.R
```diff
@@ -34,3 +34,23 @@ calc_weights <- function(net) {
...
+
+  32600 + floor((long[1] + 180) / 6) + 1
```
fnattino, line 55: I think this formula is valid only for the Northern emisphere, for the Southern emishpere it should be `32700 + ...` (suggested to rename `long` to `coords` in the previous comment as the latitude also becomes relevant):

```suggestion
  if (coords[2] >= 0.){
    base <- 32600
  } else {
    base <- 32700
  }
  
  base + floor((coords[1] + 180) / 6) + 1
```

tests/testthat/test-osm_bb.R
fnattino, line 1: Why is the test commented out? It shouldn't run anyway on CI as it is marked with `skip_on_ci()`?

fnattino (commented):
Nice @cforgaci, I have suggested a slightly different approach, but please have a look it makes sense to you!

Maybe we can add a couple of tests to make sure it works fine for all the hemispheres (I guess no need to test edge cases such as geometries crossing the date line?).

cforgaci:
> Nice @cforgaci, I have suggested a slightly different approach, but please have a look it makes sense to you!
> 
> Maybe we can add a couple of tests to make sure it works fine for all the hemispheres (I guess no need to test edge cases such as geometries crossing the date line?).

I added two tests, one for each hemisphere. Will this do?

tests/testthat/test-osm_bb.R
cforgaci: I dropped the `osm_bb()` function as it was superfluous as it was simply wrapping `osmdata::getbb()` and R CMD CHECK would not pass because of it. Fixing this issue is outside the scope of this PR though; should we do that in a separate PR?

fnattino:
Thanks @cforgaci , feel free to merge! Indeed we should fix the issue with `osm_bb` in a separate PR.

PR #26, opened 2024-09-13
Update method vignette

cforgaci:
In the first sprint, it was pointed out that the method illustration and description are not entirely clear. Also, as we are prototyping the method in R, we make some adjustments to the original method. We need to update the method description and illustration accordingly.

- [x] illustration
- [x] description


cforgaci:
I added a new illustration and updated the description accordingly. Following the logic of this diagram and description, should we have a separate valley delineation step, like the one we have for network preparation?

fnattino (approved):
Amazing @cforgaci, I love the graphics!

PR #30, opened 2024-10-07
Vignette on getting data from OSM

cforgaci:
I drafted the OSM vignette based on the notebook we had in the prototyping repository. Main changes I made:

- Code chunks with API calls are not evaluated. The vignette loads package data.
- The writing code chunks are combined into one using `purrr` in a dedicated section.
- The `leaflet` plot is replaced with a static `ggplot`.


DESCRIPTION
cforgaci, line 1: Besides adding `ggplot2` to the `Suggests` section to be able to use it in vignettes, I also ran `usethis::use_tidy_description()` which simply reorganizes the `DESCRIPTION` file according to some tidy rules.

data-raw/bucharest.R
cforgaci, line 1: There is an issue with non-ASCII characters caught by `R CMD check`. I used the solution from https://github.com/r-spatial/sf/issues/1341#issuecomment-1120902127 to fix it.

vignettes/getting-osm-data.Rmd
```diff
@@ -0,0 +1,180 @@
...
+  library(ggplot2)
+  ggplot() +
+    geom_sf(data = city_boundary, fill = "grey", color = "black") +
+    geom_sf(data = railways_lines, color = "orange") +
+    geom_sf(data = highways_lines, color = "black") +
+    geom_sf(data = waterbody, fill = "blue", color = "blue") +
+    geom_sf(data = waterway, color = "blue")
+} else {
+  message("ggplot2 not available; skipping plot examples.")
+}
```
cforgaci, line 178: As `ggplot2` is in the `Suggests` field of `DESCRIPTION`, it might not be installed on the user's system. In that case, when built locally, the vignette will display a message instead of the plot. This approach ensures that the pacakge passes `R CMD check` and the vignette is built even though `ggplot2` is not used anywhere else in the package.

fnattino: Nice solution!

vignettes/getting-osm-data.Rmd
```diff
@@ -0,0 +1,180 @@
...
+  ~ st_write(
+    .x,
+    dsn = sprintf("%s_%s.gpkg",
+                  names(st_write_list)[purrr::map_lgl(st_write_list,
+                                                      identical,
+                                                      .x)], city_name),
+    append = FALSE,
+    quiet = TRUE
+  )
+)
```
fnattino, line 161: Just to be consistent with the other libraries, maybe we can add `library("purr")` at the top and  leave the library name here:

```suggestion
walk(
  st_write_list,
  ~ st_write(
    .x,
    dsn = sprintf("%s_%s.gpkg",
                  names(st_write_list)[map_lgl(st_write_list,
                                                      identical,
                                                      .x)], city_name),
    append = FALSE,
    quiet = TRUE
  )
)
```

fnattino (commented):
Looks great @cforgaci, just a minor suggestion on my side, free to merge!

vignettes/getting-osm-data.Rmd
```diff
@@ -0,0 +1,180 @@
...
+waterbody <- bucharest$river_surface
+waterway <- bucharest$river_centerline
+```
+
+```{r packages, eval=TRUE, message=FALSE}
+library(CRiSp)
+library(dplyr)
+library(ggplot2)
+library(osmdata)
+library(sf)
```
cforgaci, line 31:
```suggestion
library(sf)
library(purrr)
```

PR #32, opened 2024-10-07
Drop traces of osm_bb() function

cforgaci:
This test was left over from `osm_bb()`. I removed that function previously and now the commented-out test with it. `R CMD check` should pass now.


fnattino:
I think both the test removal and the non-ascii character in the data will be taken care of by #30 ?

PR #33, opened 2024-10-21
Fix checks

cforgaci:
This PR fixes issues that do not pass `R CMD check`


R/retreive-osm.R
```diff
@@ -46,7 +46,7 @@ get_osmdata <- function(name, key, value) {
 get_osm_city_boundary <- function(city_name) {
   fetch_boundary <- function(key, value) {
     CRiSp::get_osmdata(city_name, key, value)$osm_multipolygons |>
-      dplyr::filter(dplyr::if_any(c(.data$`name:en`, .data$name),
+      dplyr::filter(dplyr::if_any(c(`name:en`, name),
```
fnattino, line 49: This still raises a linter warning:
```
file=R/retreive-osm.R,line=49,col=37,[object_usage_linter] no visible binding for global variable 'name:en'
```
Also a problem in rcoins, not sure what's the best approach to solve it is https://github.com/CityRiverSpaces/rcoins/issues/8

fnattino (commented):
Hi @cforgaci, looks good! Still the syntax used in `dplyr` for selection/masking seems to raise warnings with the linter. Not sure what the ideal approach is there.

cforgaci:
@fnattino, I had a look at what is changed here and I see that all issues are solved somewhere else, so I would like to close this PR without merging. Do you agree?

fnattino:
> @fnattino, I had a look at what is changed here and I see that all issues are solved somewhere else, so I would like to close this PR without merging. Do you agree?

Sure, please go ahead!

PR #34, opened 2024-10-21
Network preparation vignette

cforgaci:
I drafted the vignette for the network preparation step that needs to be carried out before delineation. I did not create CRiSp functions for this as the pre-processing steps are rather general. `insert_intersection()` and maybe some of the related utility functions we use could be suggested to be included in sfnetworks. Ideally, the code used in this vignette will be shortened and would call sfnetworks with the proposed functionality. 

@meiertgrootes, I understand from @fnattino that you came across a similar issue (inserting points where lines intersect) in the sfnetworks repo. If you can still find it, can you share a link to it? Then we can join the conversation there and submit a PR if needed.


fnattino (approved):
Looks good @cforgaci! The suggestions below are based on the linter, but feel free to discard these if you are anyway moving this to (temporary) CRiSp functions.

cforgaci:
I moved the network preprocessing code to a function `faltten_network()` and related helper functions, as well as a test checking if the correct number of linestrings is returned after cleaning. The vignette is much simpler now.

The next step would be to propose this as an additional function in the sfnetworks package. @fnattino, as you wrote this, would you like to create a PR there? Let me know how I can help with that.

PR #41, opened 2024-11-18
Implement segmentation as part of the package

fnattino:


R/delineate.R
```diff
@@ -37,20 +40,19 @@ delineate_corridor <- function(
...
+    network_filtered <- filter_network(network, corridor_buffer)
+
+    corridor <- segments(corridor, network_filtered, osm_data$river_centerline,
+                         angle_threshold)
+  }
   if (riverspace) delineate_riverspace()
   return(corridor)
 }
 
-
```
fnattino, line 45: I thought this interface, with a single access point to both the corridor and segmentation would fit better, since one actually needs the same input for both delineation (the riverspace also requires the buildings). What do you think @cforgaci ?

fnattino (commented):
Hey @cforgaci , I finally managed to move forward with the segmentation module. Happy to hear your thoughts, for now only tests are missing.

Thinks to think on (maybe later on):
* which functions should be exported and which should remain private (we should fix it consistently everyewhere)?  
* should we create a module for all the general-purpose geometry-related functions e.g. to buffer, split, find longes geometry, etc. (`geometry.R` or maybe even better `sf.R`?)

fnattino:
Rename inner functions to `get_corridor` and `get_segments`

vignettes/corridor-segmentation.Rmd
```diff
@@ -0,0 +1,49 @@
...
+)
+```
+
+```{r setup}
+library(CRiSp)
+```
+
+For a more detailed analysis of an urban river corridor, corridor-level delineation may not be sufficient. The corridor needs to be subdivided into smaller morphological units. Segmentation is a process of subdividing the corridor by using major transversal road or rail infrastructure lines.
+
+By default, the all-in-one function `delineate_corridor()` includes the division of the corridor into segments. It is also possible to use the `get_segments()` function to divide the corridor in a separate step.
```
cforgaci, line 23:
```suggestion
By default, the all-in-one function `delineate_corridor()` only returns the corridor boundary. The corridor can be segmented either by setting the argument `segments = TRUE` in `delineate_corridor()` or by using the `get_segments()` function in a separate step.
```

vignettes/corridor-segmentation.Rmd
```diff
@@ -0,0 +1,49 @@
...
+
+To demonstrate this as a separate step, we will use the `bucharest_osm$corridor` and `bucharest_osm$streets` layers from the package data as input.
```
cforgaci, line 25:
```suggestion
To demonstrate this as a separate step, we will use the `bucharest_osm$corridor`, `bucharest_osm$streets` and `bucharest_osm$railways` layers from the package data as input.
```

vignettes/corridor-segmentation.Rmd
```diff
@@ -0,0 +1,49 @@
...
+
+We first prepare the network and select all the streets and railways that cover the river corridor plus a small buffer region (see also `vignette("network-preparation")`):
+
+```{r eval=FALSE}
+# TODO remove eval=FALSE when the corridor is available as packaged data
+# Build combined street and railway network
+network <- bind_rows(streets, railways) |>
```
cforgaci, line 32:
```suggestion
network <- bind_rows(bucharest_osm$streets, bucharest_orsm$railways) |>
```

vignettes/corridor-segmentation.Rmd
```diff
@@ -0,0 +1,49 @@
...
+corridor_buffer <- sf::st_buffer(bucharest_osm$corridor, 100)
+
+# Filter the network to the area of interest
+network_filtered <- filter_network(network, corridor_buffer)
+```
+
+We then delineate segments in the corridor. The algorithm spits the corridor using river-crossing transversal edges that form continuous lines in the network:
+
+```{r eval=FALSE}
+# TODO remove eval=FALSE when the corridor is available as packaged data
```
cforgaci, line 45: Yes, this still needs to be added, but it falls outside the scope of this PR. Where should we save that? It is neither `bucharest_osm`, nor `bucharest_dem`, so it would make sense that it is a separate object called `bucharest_delineation` or `bucharest_del` which contains the `corridor`, `segments` and `riverspace`. Then we can use `bucharest_del$corridor` here. What do you think @fnattino?

cforgaci (changes requested):
@fnattino many thanks for this! I left some comments and suggestions, nothing fundamental. Can you please have a look?

cforgaci:
> * which functions should be exported and which should remain private (we should fix it consistently everyewhere)?

Maybe we expose only the top-level functions; in the `segments.r` module, for instance, we would export only `get_segments()`. We can still generate documentation for functions like `clip_and_filter()` and add `#' @noRd` to smaller helper funcions as you did. The documented functions would still be available via the `:::` namespace operator used for non-exported functions. Would that work?

> * should we create a module for all the general-purpose geometry-related functions e.g. to buffer, split, find longes geometry, etc. (`geometry.R` or maybe even better `sf.R`?)

`sf.R` would make most sense. And I see indeed that the `find_*()` functions from this module could move there.

R/segments.R
```diff
@@ -0,0 +1,245 @@
...
+#'   to form continuous strokes in the network
+#'
+#' @return Segment polygons as a simple feature geometry
+#' @export
+get_segments <- function(corridor, network, river_centerline,
+                         angle_threshold = 90) {
+
+  # Find river crossings in the network and build continuous strokes from them
+  crossings <- get_intersecting_edges(network, river_centerline)
+  crossing_strokes <- strokes(network, crossings, angle_threshold)
```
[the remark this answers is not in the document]
fnattino: I actually had a `strokes` interface function in `network.R`, but this was such a thin wrapper that it is probably not needed. Did as you suggests.

R/segments.R
```diff
@@ -0,0 +1,245 @@
...
+#' @param how Strategy for merging, choose between "smallest" (merge to smallest
+#'   adjacent block) and "longest-intersection" (merge to block which it shares
+#'   the longest intersection with)
+#'
+#' @return Blocks merged to the specified one as a simple feature geometry
+merge_block <- function(targets, block, how = "longest-intersection") {
+  idx_adjacent <- find_adjacent(targets, block)
+  if (how == "longest-intersection") {
+    intersections <- sf::st_intersection(targets[idx_adjacent], block)
+    idx_longest_intersection <- find_longest(intersections)
```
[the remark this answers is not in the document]
fnattino: I use it as short for for "index" - but I just got it is not evident enough. Renaming to `index_`

vignettes/corridor-segmentation.Rmd
fnattino: Maybe `bucharest_delineation`? Will use this here and raise an issue pointing to this thread.

fnattino:
Hi @cforgaci -  I think I have addressed all the comments, do you want to have a look?

cforgaci (approved):
Looks good @fnattino! This looks ready to be merged.

PR #50, opened 2024-11-29
Unify bbox 

fnattino:
Introduce a function `as_bbox` to standardize the bbox structure as used in `osmdata` and `valley`


cforgaci (approved):
@fnattino thanks for this, it works for me! The use of bbox was indeed clumsy.

PR #51, opened 2024-11-29
Add packaged DEM data

fnattino:
~~(builds on top of #50, merge that first)~~ Done


fnattino:
GH actions are failing because the DEM data that I have added includes invalid pointers! It turns out `SpatRaster` objects cannot be easily serialized as rds/rda files:

"These classes hold a C++ pointer to the data and they cannot be directly saved to a ".Rds" file or
used in cluster computing. They cannot be recovered from a saved R session either. See wrap or
writeRaster to work around that limitation."
https://cran.r-project.org/web/packages/terra/terra.pdf

One can bypass this by using the dedicated `saveRDS`/`readRDS` functions. Can one make these work in the current packaged data setup? If not, we might as well store the DEM as a tiff file?

fnattino:
Using `terra::saveRDS` works, but one would then "manually" call `terra::readRDS` to read the serialized object. A workaround is what I have currently implemented, which wraps the DEM into a `PackedSpatRaster` object that can be normally serialized/deserialized with the standard R package infrastructure. The drawback: one needs to call "terra::unwrap()" on the Bucharest DEM object before using it - but this is clearly stated in the representation of `PackedSpatRaster` objects:
```{r}
> CRiSp::bucharest_dem
[1] "This is a PackedSpatRaster object. Use 'terra::unwrap()' to unpack it"
```
Is this an acceptable solution?

cforgaci:
> Is this an acceptable solution?

Yes, I think this is a good solution. Thanks!

vignettes/getting-osm-data.Rmd
```diff
@@ -65,13 +65,12 @@ bucharest <- list(
   railways = railways
 )
 
-walk(
-  st_write_list,
```
fnattino, line 69: When updating the name of the dataset, I have noticed this function probably had a placeholder for the name of the dataset (`st_write_list` -> `bucharest`)? Also I made a suggestion to replace `walk` for `walk2`, which  seems to simplify quite a bit the overall snippet?

cforgaci: yes, much better!

cforgaci (approved):
prima!

PR #69, opened 2025-02-10
Improve flatten network algorithm

fnattino:
Improve the algorithm of flatten network.

I have basically only updated the loop that inject the intersections to the edge geometries, changing the order of the nested loops (first looping over edges, then over intersection points) and vectorizing operations where possible. 

Since I am now making use of the `split` base R function, I had to rename the function in `sf.R` due to name collision.

For the city of Paris, this takes the flattening from ~100 sec to ~20 sec.


cforgaci (approved):
Nice, @fnattino! For the extent I tested on Paris, flattening drops from ~23s to ~6.5s. For Bucharest, the flattening time drops from ~4s to ~2s.

Issue #64, opened 2025-01-20
Definition of the area of interest


fnattino:
We should think to a better way to define the AoI. This could be based on the river intersected with the city boundary (according to some definitions). If more segments are generated, the longest one is kept (thinking to Paris). 

One way to do this would be to take the river segment, add a buffer region, and take the bbox including this buffer.

The bbox would then be used to retrieve (network) OSM data,  and the buffer used to clip the data before setting it up as a network.

fnattino:
We should also consider a different AoI for the DEM retrieval to avoid edge effects (delineation "broadening" close to the edges).

fnattino:
Providing a little bit more context to this issue: our current approach is based on defining an area of interest (AoI) purely based on the bounding box of the city boundary, as retrieved from the OSM nominatim API from the city name:

https://github.com/CityRiverSpaces/CRiSp/blob/eb0181d60d80b866f23259fd68c1ecc3972706dd/R/delineate.R#L36

This bounding box is used to retrieve all OSM datasets (river centerline, river surface, street and rail networks) as well as the DEM for the region. This solution is not ideal, as large portions of the streets, rails, and water areas retrieved are actually not needed. Also, we realized that it would be good to increate the AoI for DEM retrieval, since the procedure of generating the valley can create edge effects close to the boundary.

So we propose to modify the approach of definition of the AoI in the following way:
* For OSM datasets, we base the AoI on the river centerline, cropped by the (bbox of the) city boundary, expanded using a given buffer (provided by input, default to 2 km?). If cropping the river leads to multiple segments, only the longest segment should be kept.
* For the DEM, we should take the bbox of the AoI defined for OSM datasets, and enlarge it by a given buffer (also provided by input, default to 1 km?).

SarahAlidoost:
in PR #76, the river centerline is cropped by the (bbox of the) city boundary, as suggested here. Why was the river centerline not cropped earlier, while the river surface was? see `get_osm_river`

fnattino:
The reason for not cropping the river centerline, was that we previously needed it to extend outside the bounding box (the intersections between the river and the bbox would lead to the start- and end-point used for the delineation). Great if we can crop it now! :)

cforgaci:
In the case of coastal (or lake-side) cities, we are encountering an additional issue: the river centerline does not extend beyond the coast. As a result, in many (if not all) of those cases, the river cannot fully cut the AoI. See the example of Rome in #85

PR #70, opened 2025-02-13
Improve on definition of corridor end points

fnattino:
With #64 we want to modify the way in which the area of interest is determined.

The AoI definition was quite critical to the workflow, since its intersections with the river centerline used to define the end points of the river corridor. Small differences to the AoI definition would thus create quite some changes in the corridor end points, and, in turn, on the corridor delineation.

Following discussions offline, this PR modifies the way in which the end points are selected. The end points are taken as the most external river crossing edges of the street (and rail) network. In this way we guarantee that the end points are always on the spatial network, hopefully making the workflow less dependent on the AoI definition. 

The approach now works ("= returns sensible end points") for cases where the river form more complex intersection with the area of interest (Paris) and for cases where the river does not fully cross the area of interest (Berlin). See few examples below (black for streets, blue for rivers, red dots are the identified corridor end-points).
 
fig-1
fig-2
fig-3
fig-4
fig-5


R/corridor.R
```diff
@@ -103,34 +109,66 @@ river_buffer <- function(river, buffer, bbox = NULL) {
...
-#'
-#' The valley is drawn on a digital elevation model (DEM) of the area,
-#' see [get_valley()] for details on the implementation.
-#'
-#' @param river A simple feature geometry representing the river
-#' @param dem The DEM of the area as a terra SpatRaster object
-#' @param bbox Bounding box defining the extent of the area of interest
-#'
-#' @return A simple feature geometry
-river_valley <- function(river, dem, bbox = NULL) {
```
fnattino, line 116: This function essentially forward a call with the same arguments to `get_valley` - I think we should drop it.

fnattino:
For some reason the linter now complains about the usage of the `return()` statement at the end of many functions. I have fixed them here, I am sorry this mixes with the actual content of this PR!

R/corridor.R
```diff
@@ -97,40 +104,72 @@ river_buffer <- function(river, buffer, bbox = NULL) {
...
+  # Only consider the edges of the spatial network
+  spat_net_edges <- sf::st_geometry(sf::st_as_sf(network, "edges"))
+
+  # Set area of interest as a polygon, even if not defined
+  if (is.null(aoi)) aoi <- sf::st_bbox(spat_net_edges)
+  aoi <- as_sfc(aoi)
+
+  # Clip the river geometry using the area of interest. This might lead to
+  # multiple river segments
+  river_segments <- sf::st_intersection(aoi, sf::st_cast(river, "LINESTRING"))
```
cforgaci, line 140: I see that you use nested functions instead of dplyr-style piped expressions. This is more compact but can be less readable with multiple function calls. We use the pipe in some other places (see, e.g., the valley module). I have a slight preference for the latter as it is more readable, even with two function calls. What do you think?

Or is the choice related to performance as well? 

This is not something to solve now (it may actually be a nice first issue for someone new in the team on my side). We can create an issue on consistency in nesting/piping multiple functions and we can add that in the contributor guidelines.

R/corridor.R
```diff
@@ -97,40 +104,72 @@ river_buffer <- function(river, buffer, bbox = NULL) {
...
+  river_network <- sfnetworks::st_network_blend(river_network, intersections)
+  nodes <- nearest_node(river_network, intersections)
+
+  # Pick a reference point on the selected river segment, either up- or down-
+  # stream from all crossings. Pick this as the closest node to the aoi boundary
+  reference <- nearest_node(river_network, sf::st_boundary(aoi))
+
+  # Compute the network distance from the reference node to all candidate end
+  # points. The closest and furthest nodes are selected as corridor end points
+  river_network <- add_weights(river_network, weight_name = "weight")
```
cforgaci, line 164:
```suggestion
  river_network <- add_weights(river_network)
```
This is the default and thus not necessary, right? Unless we use `weights = weight_name` in `sfnetworks::st_network_cost()` below and want to make the link explicit?

R/corridor.R
```diff
@@ -97,40 +104,72 @@ river_buffer <- function(river, buffer, bbox = NULL) {
...
+  river_network <- as_network(river_segments, flatten = FALSE, clean = FALSE)
+  components <- tidygraph::morph(river_network, tidygraph::to_components)
+  sum_edge_lengths <- function(x) sum(sf::st_length(sf::st_as_sf(x, "edges")))
+  edge_lengths <- sapply(components, sum_edge_lengths)
+  river_network <- components[[which.max(edge_lengths)]]
+  river_network <- clean_network(river_network)
+
+  # Find intersections with the spatial network, and add these points as nodes
+  # to the river network. These nodes are the candidates corridor end points
+  river_network_edges <- sf::st_geometry(sf::st_as_sf(river_network, "edges"))
```
cforgaci, line 153:
```suggestion
  river_network_edges <- sf::st_geometry(sfnetworks::st_as_sf(river_network, "edges"))
```
See comment above.

cforgaci (approved):
Looks good, thanks, @fnattino! I made a few small suggestions. One comment is on style and does not need to be solved in this PR.

R/corridor.R
fnattino: Correct, it is not necessary, but I indeed wanted to make the link explicit because `sfnetworks::st_network_cost` neglects weights if the `weights` input argument is not specified..

PR #74, opened 2025-02-17
Remove package name within package

### LINKED ISSUE #61: Internal functions imported using package namespace
I think one should be able to call internal functions directly, i.e. without referencing the name of the package? 

For example, the following:
https://github.com/CityRiverSpaces/CRiSp/blob/85ec2ac546a104de08fdc81c3fa09c47b3e6cd46/R/osmdata.R#L165

Could be replaced with:

```r
river_surface <- osmdata_as_sf("natural", "water", bb) 
```

Am I missing something here, @cforgaci ?

fnattino:
Closes #61


cforgaci (approved):
👍

PR #75, opened 2025-02-17
Cache download data

### LINKED ISSUE #62: Caching download data
We should add some caching mechanism on the retrieved datasets. For OSM data, maybe we should try [osmextract](https://cran.r-project.org/web/packages/osmextract/vignettes/osmextract.html) which should implement this already?

fnattino:
Closes #62


R/utils.R
```diff
@@ -93,3 +93,25 @@ reproject <- function(x, crs, ...) {
...
+  if (!is.null(bbox)) {
+    # snap spatial extent outward to include pixels crossed by the boundary
+    rasters <- lapply(rasters, terra::crop, terra::ext(bbox), snap = "out")
+  }
+  if (length(rasters) > 1) {
+    do.call(terra::merge, args = rasters)
+  } else {
+    rasters[[1]]
+  }
+}
```
fnattino, line 117: Moved this to `utils`, since its logic is quite generic

R/valley.R
```diff
@@ -38,15 +37,9 @@ get_dem <- function(bb, dem_source = "STAC", stac_endpoint = NULL,
                     stac_collection = NULL, crs = NULL) {
   bbox <- as_bbox(bb)
   if (dem_source == "STAC") {
-    if (is.null(stac_endpoint) && is.null(stac_collection)) {
```
fnattino, line 41: I have moved these defaults downstream to facilitate some testing of the cache - and I think they make more sense there.

R/osmdata.R
```diff
@@ -1,17 +1,43 @@
...
+}
+
+#' Query the Overpass API for a key:value pair within a bounding box
+#'
+#' @param key A character string with the key to filter the data
+#' @param value A character string with the value to filter the data
+#' @param bb A bounding box, in lat/lon coordinates
+#'
+#' @return An sf object with the retrieved OpenStreetMap data
+osmdata_query <- function(key, value, bb) {
```
fnattino, line 39: Separating this function makes it possible to query the Overpass API with this function without writing the cache. Useful for testing, but shouldn't be an exported function I think.

R/osmdata.R
cforgaci, line 1: For me, some tests fail locally when I try to retreive the city boundary. I added the `force_download` argument in `get_osm_city_boundary` to fix this and ran the tests with `force_download = TRUE`. I am adding the code in this comment instead of pushing the commits.

```r
get_osm_city_boundary <- function(bb, city_name,
                                  crs = NULL,
                                  multiple = FALSE,
                                  force_download = FALSE) {
  # Define a helper function to fetch the city boundary
  fetch_boundary <- function(key, value) {
    osmdata_as_sf(key, value, bb, force_download)$osm_multipolygons |>
      dplyr::filter(
        .data$`name:en` == stringr::str_extract(city_name, "^[^,]+") |
          .data$name == stringr::str_extract(city_name, "^[^,]+")
      ) |>
      sf::st_geometry()
  }
```

In `test-osmdata.R`:

```r
test_that("City boundary of Bucharest is correctly retreived", {
  skip_on_ci()
  city_name <- "Bucharest"
  bb <- get_osm_bb(city_name)
  bucharest_boundary <-
    get_osm_city_boundary(bb, city_name, force_download = TRUE)  # updated here

  expect_equal(as.numeric(bb), as.numeric(sf::st_bbox(bucharest_boundary)))
})

test_that("City boundary of Paris is returned without error", {
  skip_on_ci()

  city_name <- "Paris, France"
  bb <- get_osm_bb(city_name)
  crs <- get_utm_zone(bb)

  boundary <-
    get_osm_city_boundary(bb, city_name, crs, force_download = TRUE)  # updated here
  expect_true(!sf::st_is_empty(boundary))
})
```

cforgaci (changes requested):
Super! I pointed out two issues, one related to the use of warning() when writing to cache and the other related to a few failed osmdata tests.

R/cache.R
```diff
@@ -0,0 +1,120 @@
...
+#' @param x Object to serialize to a file
+#' @param filepath Path where to serialize x, as a character string
+#' @param wrap Whether the object should be "packed" before serialization (as
+#'   required by [`terra::SpatRaster`] objects)
+#' @param quiet Omit warning on cache file being written
+#'
+#' @return `NULL` invisibly
+write_data_to_cache <- function(x, filepath, wrap = FALSE, quiet = FALSE) {
+  if (!quiet) {
+    warning(sprintf("Saving data to cache directory: %s", filepath))
```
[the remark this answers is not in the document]
fnattino: Good point, thanks!

R/osmdata.R
fnattino: Thanks for pointing out the issue! Indeed tests on osmdata and valley, which do not run here in the CI, were a bit problematic, as they would read/write to the user cache (not ideal). I have fixed this by:
* adding a fixture that can be used to setup the temporary cache folder used for testing (see `tests/testhat/helper-cache.R`)
* porting the `force_download` argument from `osmdata_as_sf` and `load_dem` all the way downstream till `delineate`, so that user can actually configure this.

fnattino: I have also included two tests (for osmdata and DEM, respectively) which make sure the cache is properly used when downstream functions (like `get_osm_railways` and `load_dem`) are used. They are skipped on CI, so need to be tested locally.

fnattino:
Hi @cforgaci, thanks for the useful comments, I have hopefully addressed those. Can you have a final look and potentially confirm whether all tests succeed locally on your system as well?

cforgaci (approved):
Looks good. All tests pass locally.

Issue #77, opened 2025-02-19
Randomly failing R-CMD-check workflow


fnattino:
I often find the R-CMD-check workflow failing on one of the systems (either macOS, Windows or linux), on the errors similar to the following one: 

```
> base::assign(".ptime", proc.time(), pos = "CheckExEnv")
  > ### Name: get_osm_river
  > ### Title: Get the river centreline and surface from OpenStreetMap
  > ### Aliases: get_osm_river
  > 
  > ### ** Examples
  > 
  > bb <- get_osm_bb("Bucharest")
  > crs <- get_utm_zone(bb)
  > get_osm_river(bb, "Dâmbovița", crs)
  Saving data to cache directory: /Users/runner/.cache/CRiSp/osmdata_waterway_river_25.967_44.334_26.226_44.541.rds
  Error in `httr2::req_perform()`:
  ! HTTP 504 Gateway Timeout.
  Backtrace:
       ▆
    1. └─CRiSp::get_osm_river(bb, "Dâmbovița", crs)
    2.   └─CRiSp::osmdata_as_sf("natural", "water", bb, force_download = force_download) at CRiSp/R/osmdata.R:[188](https://github.com/CityRiverSpaces/CRiSp/actions/runs/13412170089/job/37464669987?pr=75#step:6:190):3
    3.     └─CRiSp:::osmdata_query(key, value, bbox) at CRiSp/R/osmdata.R:25:3
    4.       └─osmdata::osmdata_sf(...) at CRiSp/R/osmdata.R:40:3
    5.         └─osmdata:::fill_overpass_data(obj, doc, quiet = quiet)
    6.           ���─osmdata:::overpass_query(...)
    7.             └─httr2::req_perform(req)
    8.               └─httr2:::handle_resp(req, resp, error_call = error_call)
    9.                 └─httr2:::resp_abort(resp, req, body, call = error_call)
   10.                   └─rlang::abort(...)
  Execution halted
```

I think this is on running the code snippet example that is present in some of the osmdata-related docstrings for the documentation. By running the R-CMD-check workflow on multiple systems we probably fire too many requests to the Overpass API and we get frequent timeout errors like this one. Can we disable running code examples for all the functions that involve querying the Overpass API?

cforgaci:
> Can we disable running code examples for all the functions that involve querying the Overpass API?

@fnattino, would wrapping the examples into `\dontrun{}` be what is needed here?

fnattino:
[...]

PR #78, opened 2025-02-19
Refinement of corridor edges with arbitrary number of iterations

### LINKED ISSUE #68: Multiple iterations within the corridor delineation
Right now we run two iterations of the corridor delineation to get smoother corridor edges. We should make the approach more robust by setting a max number of iterations and checking for convergence while looping.

fnattino:
Closes #68.


R/corridor.R
```diff
@@ -40,17 +46,17 @@ get_corridor <- function(
...
-  # Run the delineation, setting the initial corridor geometry as excluded area
+  # Also split the network over the two regions
   network_1 <- filter_network(network, regions[1])
-  edge_1 <- corridor_edge(network_1, end_points, edges_init[1], corridor_init)
   network_2 <- filter_network(network, regions[2])
-  edge_2 <- corridor_edge(network_2, end_points, edges_init[2], corridor_init)
 
-  # Run the delineation again, with the previously determined edges as targets.
-  # This refinement step has the effect of "smoothening" the edges
-  edge_1 <- corridor_edge(network_1, end_points, edge_1)
```
fnattino, line 51: @cforgaci  In the current version of CRiSp (main branch) the initial corridor geometry (e.g. the river valley) is set as excluded area in the first iteration of the delineation, so that a penalty is added to paths that cross that region. In the second iteration, the excluded area is dropped. 

Was there some specific reason why we implemented  this way? I think I remember your run few tests on this. 

Now that I am refactoring this piece I am wondering whether we could still maintain the excluded area for paths that cross that region in the following iterations as well. 

So the code from this branch right now keeps using the excluded area for all iterations. I have tested the code snippet on Bucharest from the README file  which seems to return an almost identical delineation:

plot

What do you think?

fnattino:
Failing R-CMD-check workflows are related to #77

R/corridor.R
cforgaci: @fnattino, I think the reason was that subsequent runs should prioritise "straightening" the edge, which means that a bit of overlap with the valley would be allowed to happen. In some cases, maybe not in this one, I expect that to be useful. We can keep the generalised solution you propose until we encounter that situation.

cforgaci: I think it would be helpful to show the valley edge in this image, at least for us to assess the quality of delineation. This looks good and similar to the previous version, but I see some more deviations from the shortest path.

R/corridor.R
```diff
@@ -217,18 +223,42 @@ initial_edges <- function(corridor_initial, regions) {
 
 #' Draw a corridor edge on the spatial network.
 #'
+#' The corridor edge is drawn on the network as a shortest-path link between a
+#' start- and an end-point. The weights in the shortest-path problem are set to
+#' account for a) network edge lengths, b) distance from an initial target edge
+#' geometry, and c) an excluded area where corridor edges are aimed not to go
+#' through. The procedure is iterative, with the identified corridor edge being
```
cforgaci, line 230:
```suggestion
#' through. The procedure is iterative, with the corridor edge identified in the first iteration being
```
lintr fix needed

R/delineate.R
```diff
@@ -29,8 +31,8 @@
 delineate_corridor <- function(
   city_name, river_name, crs = NULL, bbox_buffer = NULL,
   initial_method = "valley", initial_buffer = NULL, dem = NULL,
-  capping_method = "direct", angle_threshold = 90, segments = FALSE,
-  riverspace = FALSE, ...
+  max_iterations = 5, capping_method = "direct", angle_threshold = 90,
```
cforgaci, line 34: Shouldn't `max_iterations` be passed down to `get_corridor()`?

cforgaci (changes requested):
Looks good, thanks! I left a comment and a suggestion.

R/delineate.R
fnattino: Thanks for catching this!

R/corridor.R
fnattino: I made the comparison between the corridor obtained with the corridor delineation that is currently in `main` (valley accounted as excluded area only in 1st iteration, in RED) vs this branch (valley accounted as excluded area for all iterations, in BLUE). The valley is added as the ORANGE polygon. 

Probably you are right in that using excluded area only in the first iteration leads to less deviations in the edges, so I would say, let's go back to the same approach used in `main`? What do you think?

Bucharest-compare-excluded-area

cforgaci: I agree we go back to the version on `main`. Indeed the extra deviations seem to occur due to the overlap penalty.

cforgaci:
@fnattino, any ideas why the overlap in the circled area happens in the first place? I would expect the delineation to follow the green line.

415592183-b82c3a13-eec7-493e-9aa0-81e7878a65c2

This may be out of scope in this PR, so feel free to merge if you think we should address this in a separate issue. Apart from this comment, I think this is ready to be merged.

cforgaci (approved):
See my previous comment as well.

fnattino:
> @fnattino, any ideas why the overlap in the circled area happens in the first place? I would expect the delineation to follow the green line.

Not really sure where this is coming from. This is obtained if I plot the edges obtained for all iterations (all in RED). As you can see, the deviation from your "ideal" corridor geometry, is already there from the first iteration (there are no RED lines that follow that path).

Bucharest

fnattino:
> > @fnattino, any ideas why the overlap in the circled area happens in the first place? I would expect the delineation to follow the green line.
> 
> Not really sure where this is coming from. This is obtained if I plot the edges obtained for all iterations (all in RED). As you can see, the deviation from your "ideal" corridor geometry, is already there from the first iteration (there are no RED lines that follow that path).
> 

I have also tried to either drop the excluded area completely or to increase the penalty parameter by 10 times. Nothing changes in that segment of the corridor (other parts of the corridor get significantly worse).

cforgaci:
> Not really sure where this is coming from. This is obtained if I plot the edges obtained for all iterations (all in RED). As you can see, the deviation from your "ideal" corridor geometry, is already there from the first iteration (there are no RED lines that follow that path).

Exactly, that is what I meant when I said that this might be out of scope in this PR. I was just wondering why such deviation occurs in the first iteration if it overlaps so much with the valley and there is an apparent shorter path where I drew the green line.

cforgaci:
> > Not really sure where this is coming from. This is obtained if I plot the edges obtained for all iterations (all in RED). As you can see, the deviation from your "ideal" corridor geometry, is already there from the first iteration (there are no RED lines that follow that path).
> 
> Exactly, that is what I meant when I said that this might be out of scope in this PR. I was just wondering why such deviation occurs in the first iteration if it overlaps so much with the valley and there is an apparent shorter path where I drew the green line.

I guess. this might be due to the edge effect from the underlying DEM, so maybe increasing the extent of it with a few kilometers (say, something like the average width of the valley), without changing the penalty parameter, might lead to an improved result. This can be addressed in another PR.

fnattino:
> I guess this might be due to the edge effect from the underlying DEM, so maybe increasing the extent of it with a few kilometers (say, something like the average width of the valley), without changing the penalty parameter, might lead to an improved result. This can be addressed in another PR.

Yes, using a DEM on a larger area of interest than actually required will be dealt within #76!

PR #87, opened 2025-03-04
set aws env variable for unsigned requests

### LINKED ISSUE #46: Credentials to access COP DEM on AWS?
Are credentials needed in order to access the DEM tile? The STAC catalog includes  `s3://` links as assets, not sure whether terra (GDAL) can access these without credentials, even if data is in a free bucket

SarahAlidoost:

- [ ] Refactor
- [ ] Feature
- [X] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related Issues

- Closes #46

## Added/updated tests?

- [ ] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes

### DIFF
--- a/R/valley.R
+++ b/R/valley.R
@@ -102,9 +102,14 @@ get_stac_asset_urls <- function(bb, endpoint = NULL, collection = NULL) {
   if (is.null(endpoint) && is.null(collection)) {
     endpoint <- default_stac_dem$endpoint
     collection <- default_stac_dem$collection
+    # check if there is AWS credentials file, otherwise use unsigned requests
+    if (!file.exists("~/.aws/credentials")) {
+      Sys.setenv("AWS_NO_SIGN_REQUEST" = "YES")
+    }
   } else if (is.null(endpoint) || is.null(collection)) {
     stop("Provide both or neither of STAC endpoint and collection")
   }
+
   bbox <- as_bbox(bb)
   rstac::stac(endpoint) |>
     rstac::stac_search(collections = collection, bbox = bbox) |>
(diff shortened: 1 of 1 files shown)


fnattino (changes requested):
Great this worked @SarahAlidoost! Just a minor comment: I think the access issue is with the raster data files, not with the STAC catalog, which should always be accessible. So I would move the lines you added to the `load_dem` function, somewhere before the call to `load_raster` - then good to merge for my side!

SarahAlidoost:
> Great this worked @SarahAlidoost! Just a minor comment: I think the access issue is with the raster data files, not with the STAC catalog, which should always be accessible. So I would move the lines you added to the `load_dem` function, somewhere before the call to `load_raster` - then good to merge for my side!

thanks for the review. I noticed it, but the reason is that `endpoint` and `collection` are set to default (AWS) in `get_stac_asset_urls` function. If `endpoint` and `collection` are not related to AWS, another env variable should be set. In case of setting AWS env variable in `load_dem` as you suggested, I think the `tile_url` should be inspected to see if it includes "AWS".  Right? If so, I think adding them to `get_stac_asset_urls` is a nicer solution. Please let me know if I am missing something.

fnattino:
Good point, I haven't thought to this so thoroughly. 

Correct, the problem in accessing data would be there for any dataset on AWS buckets, so I agree it would be better to check the URLs. But rather than checking whether AWS is is the URLs, I would check whether (one of) the URLs starts with the `s3://` protocol. 

I also agree that in this sense this can also fit in `get_stac_asset_urls` - this is some sort of environment preparation to access the data. So I am happy to leave it there!

PR #92, opened 2025-03-11
Add river centerline to vpoints source geometry

### LINKED ISSUE #91: Viewpoints from `river_centerline` when `river_surface` is (partially) missing
In some cities river surface OSM data is missing, either partially or entirely. This is an issue for riverspace delineation, because at the moment viewpoints are only created from the river surface polygon.

So, whenever the river surface polygon is unavailable, we should use the river centerline, which is always available, as a fallback geometry. This should be done with a warning because rays are traced with the same radius regardless of the width of the river. The warning should also specify when the geometry which the viewpoints are based on is a combination of river centerline and partial river surfaces.

At least for now, the delineation should not discriminate between above-ground and underground river segments.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
With this PR, `river_centerline` is also used for generating viewpoints.

## Related Issues

- Related Issue #91 
- Closes #91 

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


R/utils.R
```diff
@@ -139,3 +139,17 @@ load_raster <- function(urlpaths, bbox = NULL) {
     rasters[[1]]
   }
 }
+
+#' Combine river centerline and surface
+#'
+#' @param river_centerline River line as sfc_LINESTRING or sfc_MULTILINESTRING
+#' @param river_surface River surface as sfc_POLYGON or sfc_MULTIPOLYGON
+#'
+#' @return Combined river as sfc_LINESTRING
```
fnattino, line 148: Isn't this a multilinestring instead?
```suggestion
#' @return Combined river as sfc_MULTILINESTRING
```

fnattino (changes requested):
Looks good to me, maybe add one of a few tests?

R/utils.R
cforgaci: It is, indeed!

cforgaci:
> Looks good to me, maybe add one of a few tests?

@fnattino, I added a few tests. Please have another look.

fnattino (approved):
Great, thanks!

Issue #67, opened 2025-01-20
River and city names in OSM queries


fnattino:
We should find a way to partially match river and city names as returned by osmdata queries.

fnattino:
[Changes here](https://github.com/CityRiverSpaces/CRiSp/pull/96#discussion_r1991762259) as part of #96 may actually be checking multiple languages - we should test this, and whether this also covers partial matches. We should also integrate this with the similar filtering mechanism used for the city boundary.

fnattino:
We could move what it is currently done for the rivers:
```r
    dplyr::filter(dplyr::if_any(dplyr::matches("name"),
                                \(x) x == river_name))
```
to a function, and use this both for the city boundaries and for the rivers

fnattino:
#114 allows river name and city name to be passed in any language, but they still need to be exact match of the OSM values. Should we also add partial matches?

cforgaci:
> #114 allows river name and city name to be passed in any language, but they still need to be exact match of the OSM values. Should we also add partial matches? 

@fnattino thanks for taking over. One of the reasons why #114 was still open is because I was still trying to figure out partial matching.

The thing is that Nominatim already takes care of partial matching or misspelling. See for instance https://nominatim.openstreetmap.org/ui/search.html?q=Dambovita

Doesn't osmdata allow for that? Maybe there is a default parameter that disables that? Just thinking... maybe you already checked.

PR #96, opened 2025-03-12
Fix corridor delineation and segmentation 

fnattino:
This PR introduces few changes that allow the corridor delineation and segmentation to succeed (i.e. succesfully complete) on all the cities considered in https://github.com/CityRiverSpaces/CRiSp-demo-material/blob/main/scripts/generate-corridor.R (plus Paris and Berlin).

As main change: for the corridor delineation, we relied so far on the river centerline splitting a the area of interest (AoI) in two parts to define the two river banks where to look for edges. This approach was not robust enough, and even trying to build the AoI built as a "flat" buffer region around the same river did not work fine due to numerical reasons. Here, I suggest to use instead single-sided buffers build from the river centerline to bypass the issue.

### DIFF
--- a/R/corridor.R
+++ b/R/corridor.R
@@ -9,6 +9,8 @@
 #'   centerline
 #' @param river_surface A simple feature geometry representing the river surface
 #' @param aoi Area of interest as sf object or bbox
+#' @param max_width (Approximate) maximum width of the regions considered on the
+#'   two river banks
 #' @param initial_method The method employed to define the initial river
 #'   corridor geometry. See [initial_corridor()] for the available methods
 #' @param buffer Buffer region to add to the river geometry to setup the initial
@@ -24,8 +26,9 @@
 #' @return A simple feature geometry representing the river corridor
 #' @export
 delineate_corridor <- function(
-  network, river_centerline, river_surface, aoi, initial_method = "valley",
-  buffer = NULL, dem = NULL, max_iterations = 10, capping_method = "direct"
+  network, river_centerline, river_surface, aoi = NULL, max_width = 2500,
+  initial_method = "valley", buffer = NULL, dem = NULL, max_iterations = 10,
+  capping_method = "direct"
 ) {
   # Drop all attributes of river centerline and surface but the geometries
   river_centerline <- sf::st_geometry(river_centerline)
@@ -37,11 +40,15 @@ delineate_corridor <- function(
                                     buffer = buffer, dem = dem,
                                     bbox = as_bbox(aoi))
 
-  # Pick the corridor end points
-  end_points <- corridor_end_points(river_centerline, network, aoi = aoi)
+  # Build river network in the defined area of interest
+  river_network <- build_river_network(river_centerline, aoi = aoi)
 
-  # Split the area of interest along the river centerline
-  regions <- split_aoi(aoi, river_centerline)
+  # Pick the corridor end points as the two furthest crossings between the
+  # spatial network and the river
+  end_points <- corridor_end_points(river_network, network)
+
+  # Define the spatial regions corresponding to the two river banks
+  regions <- get_river_banks(river_network, max_width)
 
   # Determine the initial corridor edges by splitting the initial corridor
   # boundary through the just determined regions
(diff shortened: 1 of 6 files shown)


R/delineate.R
```diff
@@ -74,27 +74,31 @@ delineate <- function(
     force_download = force_download
   )
 
-  # Get the bounding box and (if not provided) the CRS
-  if (is.null(crs)) crs <- get_utm_zone(osm_data$aoi)
+  # If not provided, determine the CRS
+  if (is.null(crs)) crs <- get_utm_zone(osm_data$bb)
```
fnattino, line 78: The `aoi` here could not be present if corridor is FALSE. We can re-add the `bb` to be returned by `get_osm_data`, shouldn't hurt.

R/osmdata.R
```diff
@@ -104,28 +104,37 @@ get_osmdata <- function(
...
     boundary = boundary,
     river_centerline = river$centerline,
     river_surface = river$surface
   )
 
   # Retrieve streets and railways based on the aoi
   if (!is.null(network_buffer)) {
-    aoi <- get_river_aoi(river, bb, buffer_distance = network_buffer)
-    osm_data <- append(osm_data, list(aoi = aoi))
+    aoi_network <- get_river_aoi(river$centerline, bb,
```
fnattino, line 115: For consistency, I thought it would be good to actually add the AoI used to retrieve buildings as well (if needed). Hence the renaming to be more specific on this being only for the network.

R/osmdata.R
```diff
@@ -217,10 +226,13 @@ get_osm_river <- function(bb, river_name, crs = NULL, force_download = FALSE) {
   river_centerline <- osmdata_as_sf("waterway", "river", bb,
                                     force_download = force_download)
   river_centerline <- river_centerline$osm_multilines |>
-    dplyr::filter(.data$name == river_name) |>
+    # filter using any of the "name" columns (matching different languages)
```
fnattino, line 229: This should allow for checking multiple columns with `name` in as header (i.e. multiple languages) - we could integrate this and the similar filtering used for the city boundary in a separate function, potentially including partial name matching. This should help with #67

R/network.R
```diff
@@ -142,24 +142,26 @@ calc_rolling_sum <- function(x, n = 2) {
...
 # nolint start
 #' Subdivide edges by [adding missing nodes](https://luukvdmeer.github.io/sfnetworks/articles/sfn02_preprocess_clean.html#subdivide-edges),
-#' simplify the network (see [`simplify_network()`]), remove
+#' (optionally) simplify the network (see [`simplify_network()`]), remove
 #' [pseudo-nodes](https://luukvdmeer.github.io/sfnetworks/articles/sfn02_preprocess_clean.html#smooth-pseudo-nodes),
 #' and discard all but the main connected component.
 # nolint end
 #'
 #' @param network A network object
+#' @param simplify Whether
```
cforgaci, line 151:
```suggestion
#' @param simplify Whether the network should be simplified
```

R/corridor.R
```diff
@@ -57,8 +64,12 @@ delineate_corridor <- function(
   edge_2 <- corridor_edge(network_2, end_points, edges_init[2], corridor_init,
                           max_iterations)
 
-  # Cap the corritor and return it
-  cap_corridor(c(edge_1, edge_2), capping_method, network)
+  # Cap the corritor
+  corridor <- cap_corridor(c(edge_1, edge_2), capping_method, network)
+
+  # Loops along the edges (or capping segments) form additional polygons, that
+  # can be discarded by selecting only the largest feature
```
cforgaci, line 71: It is not clear to me what "loops along the edges (or capping segments) form additional polygons" are

R/corridor.R
```diff
@@ -99,34 +110,27 @@ initial_corridor <- function(
   }
 }
 
-#' Find the corridor end points.
+#' Build a spatial network from the river center spatial features
```
cforgaci, line 113:
```suggestion
#' Build a spatial network from river centerlines
```

R/corridor.R
```diff
@@ -135,56 +139,74 @@ corridor_end_points <- function(river, network, aoi = NULL) {
...
-#' @param bbox Bounding box defining the extent of the area of interest
-#' @param river A simple feature geometry representing the river centerline
-#'
-#' @return A simple feature geometry set of two areas of interest
-#' @importFrom rlang .data
-split_aoi <- function(bbox, river) {
-  regions <- split_by(as_sfc(bbox), river)
+#' @param river River spatial features provided as a [`sfnetworks::sfnetwork`]
+#'   or [`sf::sf`]/[`sf::sfc`] object.
+#' @param width (Approximate) width of the regions
```
cforgaci, line 192: Why approximate? I am just curious.

cforgaci (approved):
@fnattino, nice approach with the river network and side buffers for aoi! I left a few comments, but it can be merged anytime.

SarahAlidoost (changes requested):
@fnattino thanks for the fixes, I noticed that this PR will change the corridor of the example in readme, see below. You might update the figure in this PR or check if the changes make sense. 
image

R/network.R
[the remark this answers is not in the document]
fnattino: Uh 🫣 - thanks!

R/corridor.R
fnattino: Actually, this the "real" width - it  is`max_width` in `delineate_corridor` to be an approximate indication of the maximum width of the corridor, since the actual corridor can extend slightly outside the region if the edges of the network "spill" outside it. I will the text here, thanks for checking this scrupulously!

R/corridor.R
fnattino: I agree with you - when re-reading this now it sounds pretty cryptic 😅 .  What I noticed for a couple of cities is that the corridor edges would cross with one of the "capping" edges forming a sort of a loop-hole on the corridor boundary. When polygonized, this would lead to a small side polygon at the end of the corridor, which we ideally want to remove. Will try to tackle this better here.

fnattino:
> @fnattino thanks for the fixes, I noticed that this PR will change the corridor of the example in readme, see below, the edges. You might update the figure in this PR or check if the changes make sense.

Thanks @SarahAlidoost, I have update the figure!

R/corridor.R
```diff
@@ -283,5 +301,17 @@ cap_corridor <- function(edges, method = "direct", network = NULL) {
...
+  # than the end points, the polygonization of the corridor boundary leads to
+  # small side polygons. We drop these, after raising a warning
+  if (length(polygon) > 1) {
+    warning(
+      "Corridor capping gives multiple polygons - selecting the largest one"
+    )
+    polygon <- polygon[find_largest(polygon)]
+  }
+
+  polygon
```
fnattino, line 316: @cforgaci  Following up on https://github.com/CityRiverSpaces/CRiSp/pull/96#discussion_r1995857151 , do you this is clear now? Moved it inside `cap_corridor` since this is where the problem can actually arise.

cforgaci: Yes, super clear now!

PR #110, opened 2025-03-20
Swap OSM tags when retrieving city boundary

### LINKED ISSUE #83: Correct OSM city boundary retreival
I see that `get_osm_city_bounday()` still returns no or wrong boundaries quite often.

See this example for Bratislava, for instance:

Image
 
We need to make this function more robust. A few things to consider:
- Partial or even mistyped city names should still match the correct city name
- If only one boundary is found, no need to use the filter in the helper function in `fetch_bounday()`
- The currently used `key=value` pairs may not cover all possibilities

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
With this PR, `get_osm_city_boundary` first tries to get the data first with the tag `boundary:administrative` and then `place:city`. Most city boundaries are recorded (correctly) with the former, so it makes sense to begin with that one. Previously it was the other way around and would return wrong boundaries, like in the case of Bratislava. 

## Related Issues

- Related Issue #83 
- Closes #83

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


fnattino (commented):
Thanks @cforgaci. Just a curiosity: are there any cities for which `boundary:administrative` does not work and `place:city` does? If this is not the case we could simply drop the search for  the `place:city` tag ..

fnattino:
It turns out the size of the boundary/administrative query is quite large - do we still want to invert the order?

fnattino:
Sounds good to only keep `boundary:administrative` - but maybe let's make the city boundary retrieval optional within `get_osmdata()`)? 

My point is that we could have `get_osmdata()` retrieving by default the city boundary, but within the top-level function `delineate()` we should avoid retrieving data that is not actually used (and not returned to the user either). So, within `delineate()`, we could do something like:
```r
osmdata <- get_osmdata(..., city_boundary = FALSE, ...)
```

cforgaci:
@fnattino, I updated `get_osm_city_boundary()` to only use `boundary:administrative` and I disabled city_boundary retrieval in `delineate()` as you suggested. Will this work?

fnattino (commented):
Looks good @cforgaci, just one suggestion, then good to merge on  my side!

PR #112, opened 2025-03-21
Fix invalid buildings retrieved from OSM

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
When trying to run riverspace delineation on `"Köln", "Rhein"`, I encountered invalid building polygons. This change in PR ensures that the polygons are made valid before being intersected with the aoi.

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes

### DIFF
--- a/R/osmdata.R
+++ b/R/osmdata.R
@@ -346,6 +346,8 @@ get_osm_buildings <- function(aoi, crs = NULL, force_download = FALSE) {
   buildings <- osmdata_as_sf("building", "", aoi,
                              force_download = force_download)
   buildings <- buildings$osm_polygons |>
+    # retreived buildings may contain invalid polygons, fix these
+    sf::st_make_valid() |>
     sf::st_filter(aoi, .predicate = sf::st_intersects) |>
     dplyr::filter(.data$building != "NULL") |>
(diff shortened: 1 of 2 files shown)


fnattino (approved):
Thanks @cforgaci  for raising this, I think we can merge it.

There were similar problems with the water surface, and I think the problem arises from having small polygons in lat/lon coordinates that are invalid when considering the spherical space (they are fine if assuming a projected space though). So another solution for both this issue and the water surface one would be to work with the geographical coordinates as if they were projected ones, which can be done by setting `sf_use_s2(FALSE)` (the `GEOS` backed is then used for both geographical and projected coordinates). Considering the rather limited extent of our areas of interest, and the fact that we are then anyway mostly working with projected coords, I think this should be fine. Dropping the `s2` backend seems to also give some performance benefits..  

So maybe let's merge this, and open an issue on testing whether we can instead keep the polygons as they are retrieved by OSM by setting `sf_use_s2(FALSE)`?

Issue #126, opened 2025-03-25
Reprojection of bounding box failing on Windows


fnattino:
On the Windows  R-CMD-check CI we get the following failing error (see [action](https://github.com/CityRiverSpaces/CRiSp/actions/runs/14058747081/job/39375384719)):

```
 ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Error ('test-utils.R:188:3'): reproject works with bbox ─────────────────────
  Error in `st_transform.bbox(x, crs, ...)`: method not available for GDAL: 3.10.1
  Backtrace:
      ▆
   1. └─CRiSp::reproject(x, 32602) at test-utils.R:188:3
   2.   ├─sf::st_transform(x, crs, ...) at CRiSp/R/utils.R:133:5
   3.   └─sf:::st_transform.bbox(x, crs, ...)
```

fnattino:
Seems like the problem is not related to windows, but to the version of GDAL - opened an issue on the topic on `sf` https://github.com/r-spatial/sf/issues/2512

fnattino:
A fix has already been put in!

cforgaci:
Nice! Does this mean that the fix is available in the dev version of sf while the gh action uses the CRAN version?

fnattino:
Correct, the fix is now in the main branch on the sf GH repo, it won't fix our failing GH action until a new sf version is released on CRAN. Fix that we could (temporarily) by replacing our dependency with the `sf` dev version (but very much undesirable, since we would need to build sf from source for every installation, with all problems related to system dependencies).

PR #127, opened 2025-03-25
Fix interface to visor fn

fnattino:
Fixing interface to visor, to test after https://github.com/CityRiverSpaces/visor/pull/5 is merged


fnattino:
The failing action is again related to #126 , so we can ignore it for now. 

I have updated the default number of rays to a multiple of 4, since the number of rays per quadrant need to be a whole number. [This PR](https://github.com/CityRiverSpaces/visor/pull/7) in visor will make the isovist calculation more robust with respect to the initial value of `ray_num`.

cforgaci (approved):
👍

PR #131, opened 2025-03-26
Fix for function without examples

fnattino:
Fix to solve issue in package check: https://github.com/CityRiverSpaces/CRiSp/issues/10#issuecomment-2753592375


cforgaci (approved):
👍

PR #132, opened 2025-03-26
Fix segmentation bug

### LINKED ISSUE #129: BUG found in segmentation of Nile corridor in Cairo
Problem identified here:

https://github.com/CityRiverSpaces/CRiSp-demo-material/pull/6#pullrequestreview-2710680091

It turns out that `split_by` in `clip_and_filter` returns three segments, and the wrong ones are used to determine which network strokes are fully-crossing the corridor. Probably the two longest ones should be returned? 

https://github.com/CityRiverSpaces/CRiSp/blob/c157d71d1b165fb1923699fea8568d01c4ebff57/R/segments.R#L57C3-L57C74

fnattino:
Closes #129 

Segmentation of the Nile river corridor in Cairo now returns multiple segments:

Screenshot 2025-03-26 at 12 45 24


cforgaci (approved):
Nice fix. Thanks.

The result of segmentation raises some questions (as it does in some other cases), but those are definitely outside the scope of this PR and can be addressed later.

PR #141, opened 2025-03-31
Different approach to segmentation

fnattino:
This PR proposes a slightly different approach to segmentation.

Currently we identify the strokes in the spatial network that crosse the river center line and both corridor edges and use these to "split" the corridor into subregions, which are the candidate river segments. These regions are then "post-processed" to make sure valid river segments are returned.

This PR replaces this approach with an approach that is instead based on selecting proper strokes from the network: we start by dropping candidate segmentation edges that form most intersections with others. For equal number of intersections, we drop the longest line.


fnattino:
Here is the comparison of the original segmentation (from `main` branch) and the current delineation (this branch)

| Original Segmentation | Current Segmentation |
|----------------------|------------------------|
| Berlin | Berlin |
| Bristol | Bristol |
| Bucharest | Bucharest |
| Cairo | Cairo |
| Dresden | Dresden |
| Graz | Graz |
| Izmir | Izmir |
| Köln | Köln |
| London | London |
| Los Angeles | Los Angeles |
| Madrid | Madrid |
| Prague | Prague |
| Warsaw | Warsaw |

R/delineate.R
```diff
@@ -43,7 +43,7 @@ delineate <- function(
   city_name, river_name, crs = NULL, network_buffer = NULL,
   buildings_buffer = NULL, dem_buffer = 2500, initial_method = "valley",
   initial_buffer = NULL, dem = NULL, max_iterations = 10,
-  capping_method = "shortest-path", angle_threshold = 90, corridor = TRUE,
+  capping_method = "shortest-path", angle_threshold = 100, corridor = TRUE,
```
cforgaci, line 46: Why has the default angle threshold been increased to 100?

R/segments.R
```diff
@@ -24,30 +24,29 @@
...
+                               angle_threshold = 100) {
 
   # Find river crossings in the network and build continuous strokes from them
   crossings <- get_intersecting_edges(network, river_centerline, index = TRUE)
   crossing_strokes <- rcoins::stroke(network, from_edge = crossings,
                                      angle_threshold = angle_threshold)
 
-  # Clip strokes and select the ones that could be used as segment boundaries
-  block_edges <- clip_and_filter(crossing_strokes, corridor, river_centerline)
+  # Clip strokes to the corridor extend and select non-intersecting strokes as
```
cforgaci, line 34:
```suggestion
  # Clip strokes to the corridor extent and select non-intersecting strokes as
```

cforgaci (approved):
Great alternative approach! I left a tiny suggestions and a question.

R/delineate.R
fnattino: Good that you point this out. I noticed results looks slightly better for a higher threshold. I suppose this is because earlier we would have the post-processing step that would still refine the segments, so we could be a little bit more "loose" on what we actually consider as acceptable continuous straight lines.  Right now, the lines generated by rcoins are the lines that are ultimately used as segment boundaries, so to make sure these are "straight enough" we have to apply a higher threshold. Does this make sense to you?

cforgaci: Yes, thanks for the explanation!

PR #142, opened 2025-03-31
add check for invalid geometries before st_filter

### LINKED ISSUE #115: Use of the S2 back-end results in invalid geometries when rertrieving buildings and water surfaces
There were similar problems with the water surface, and I think the problem arises from having small polygons in lat/lon coordinates that are invalid when considering the spherical space (they are fine if assuming a projected space though). So another solution for both this issue and the water surface one would be to work with the geographical coordinates as if they were projected ones, which can be done by setting `sf_use_s2(FALSE)` (the `GEOS` backed is then used for both geographical and projected coordinates). Considering the rather limited extent of our areas of interest, and the fact that we are then anyway mostly working with projected coords, I think this should be fine. Dropping the `s2` backend seems to also give some performance benefits..  

So maybe let's merge this, and open an issue on testing whether we can instead keep the polygons as they are retrieved by OSM by setting `sf_use_s2(FALSE)`?

_Originally posted by @fnattino in https://github.com/CityRiverSpaces/CRiSp/pull/112#pullrequestreview-2709863924_

SarahAlidoost:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [X] Optimization
- [ ] Documentation Update

## Related Issues

- See [this comment](https://github.com/CityRiverSpaces/CRiSp/issues/115#issuecomment-2765664961)
- Closes #115 

## Added/updated tests?

- [ ] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


fnattino:
@SarahAlidoost  just a heads up that the Windows R-CMD-check is broken because of #126, so we currently expect it to fail

fnattino (changes requested):
Thanks a lot @SarahAlidoost for this! As I tried to explain in #115, I think that even working with `GEOS`  vs `s2` would not help as invalid polygons would remain as such. 
So I think we really have to use `sf::st_make_valid()`. 

Maybe this PR could still be useful as it is good to receive the warning if some of the retrieved geometries  are invalid. I would however drop the `sf::st_use_s2()` and re-add instead a call to `sf::st_make_valid()`..

fnattino (approved):
Sorry for the late reply - this is good to go, thanks @SarahAlidoost!

Issue #180, opened 2025-04-14
Missing harbours/water surface in Rotterdam and Arnhem


fnattino:
By keeping the only water areas that intersect the river centerline, we miss quite some important water areas like harbours in Rotterdam and Arnhem.

cforgaci:
This seems to be the same issue as described in #161

cforgaci:
Closing this duplicate

Issue #161, opened 2025-04-12
Harbor missing from water surface in Bratislava


cforgaci:
Danube harbors in Bratislava are not returned with river surface.

cforgaci:
I checked and those polygons are dropped because they have a different name. See this one, for instance:

![Image](https://github.com/user-attachments/assets/162e0c27-1fa1-4e60-bb07-28e35067cc97)

The quesiton is, can we select touching polygons like that or do we accept that they will be dropped. Note that polygons like the ones in  #175 should be still dropped/clipped.

cforgaci:
Currently, the harbor highlighted above is returned, but due to the coincidence that it intersects "Little Danube" which is currently retrieved along with "Danube" because of the way we match names. Note that the other harbor, namely Lodenica, is not retrieved.

Now, this is more of a use case issue, and only for river space delineation, as pointed out in #175. I am closing the issue here and created https://github.com/CityRiverSpaces/rcrisp-usecases/issues/12 instead.

PR #184, opened 2025-04-15
Remove river surface from corridor delineation

### LINKED ISSUE #103: Make river surface optional
Some of the rivers we have considered do not have `natural:water` OSM features, so the `river_surface` remains an empty collection.  Corridor and delineation and segmentation seems to work fine even in these cases, and the riverspace will be fixed by #92 (issue #91). We should however make the `river_surface` parameter optional e.g. in `delineate_corridor`, so that the function can also be used if this is not available.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
This removes river surface from corridor delineation. Only the river centerline is used now.

## Related Issues

- Related Issue #103 #180 #161 
- Closes #103 

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: I am not sure if an additional test is needed here

## Added entry in changelog?

- [ ] Yes


fnattino (approved):
Looks good to me! Should we try to run the example in the README to see whether this actually gives any differences, at least for Bucharest?

fnattino:
Indeed, looks almost the same - just wanted to make sure that this would not cause drastic changes.  I think we can merge!

PR #187, opened 2025-04-25
Drop `aoi` from corridor delineation

### LINKED ISSUE #149: Remove `aoi` from delineate_corridor
Can we actually simplify delineate corridor so that the `aoi` can be removed from the input arguments? The extent of the DEM and of the network should do it, so I think there should be no need for the additional info.

fnattino:
Closes #149


fnattino:
I have checked and it does not affect the cities considered e.g. in #141 .

cforgaci (approved):
Looks good. Sharpens corridor delineation and related functions. Thanks!

PR #188, opened 2025-04-28
Increase default value of network buffer

fnattino:
For some cases (e.g. Warsaw and Rome), the default value of network buffer is too small - we increase this from 2500m to 3000m

### DIFF
--- a/R/delineate.R
+++ b/R/delineate.R
@@ -6,7 +6,7 @@
 #'   provided, the suitable Universal Transverse Mercator (UTM) CRS is selected
 #' @param network_buffer Add a buffer (an integer in meters) around
 #'   river to retrieve additional data (streets, railways, etc.).
-#'   Default is 2500 m.
+#'   Default is 3000 m.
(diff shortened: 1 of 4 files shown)


R/delineate.R
```diff
@@ -6,14 +6,14 @@
...
-#'   Default is 2500 m.
+#'   Default is 3000 m.
 #' @param buildings_buffer Add a buffer (an integer in meters) around the
 #'   river to retrieve additional data (buildings). Default is 100 m.
 #' @param dem_buffer Add a buffer (an integer in meters) for retrieving the DEM.
 #' @param initial_method The method employed to define the initial river
 #'   corridor geometry. See [initial_corridor()] for the available methods
 #' @param dem_buffer Add a buffer (an integer in meters) around the
-#'   river to retrieve the DEM. Default is 2500 m.
+#'   river to retrieve the DEM.
```
fnattino, line 16: For `dem_buffer` the default value is already shown by the function signature.

cforgaci (approved):
👍 thanks

Issue #181, opened 2025-04-14
Corridor artifacts in Vistula/Warsaw


fnattino:
For Warsaw, the corridor has a weird artifact on the South East side, to be checked. Also it does seem to run on a line that is too external on the left side. 
 
![Image](https://github.com/user-attachments/assets/d8ff5bef-3bc9-49c9-aa27-f6f39cdedef8)

fnattino:
The problem with the corridor can be resolved by increasing the network buffer to 3000m - considering that for other cases as well (Rome, Madrid) the default value of 2500m is too narrow will update the default in a PR. 
![Image](https://github.com/user-attachments/assets/f8c26885-9b8d-445e-b775-69c8081f9a6f)

PR #189, opened 2025-04-29
Improve corridor end point selection

### LINKED ISSUE #173: Strange corridor capping in some cases
The capping in the norther end of the Rio Manzanares, Madrid corridor is only on one side of the river. This is probably related to the size of the AoI #174

fnattino:
This PR solves issues related to the corridor end point selections such as the ones illustrated in #181  and #173 (closes #173). 

While we earlier only checked for crossings between the river and the spatial network, we now make sure that these crossings connect the two sub-networks.


fnattino:
I have tested few cities and I got identical results for all but for Berlin and Madrid, for which we see improvements in the capping close to the end points of the corridor:

| `main` | current |
|--------|--------|
| Berlin | Berlin |
| Madrid | Madrid |

R/network.R
```diff
@@ -339,20 +339,30 @@ nearest_node <- function(network, target) {
   nodes[idx]
 }
 
-#' Subset a network keeping the only nodes that intersect a target geometry.
+#' Subset a network keeping the components that intersect a target geometry.
```
cforgaci, line 342: Maybe a different word than "component" to avoid confusion with (connected) components as connected regions/sub-graphs in a network? Maybe we can call them "elements" instead?

cforgaci (commented):
Thanks! I left one comment on the naming of a parameter.

R/network.R
fnattino: You are right, it is probably better to stick to the tidigraph naming of "component" as sub-graph. Will adapt to use "elements" instead!

PR #190, opened 2025-04-29
Initialise exampledata module

### LINKED ISSUE #185: Package data cannot be published on CRAN
I am pasting here the last part of our conversation about the package data:

@fnattino:
> OK, so basically the approach with a data package on CRAN and multiple licenses won’t work.
>  
> Possible approaches to bypass the issue that I can think of:
> * We keep CRiSpData as it is, but we avoid publishing it to CRAN. We can try to submit CRiSp to CRiSp with CRiSpData being a “remote” soft dependency, or remove CRiSpData entirely from dependencies and add checks on CRiSpData being installed before performing tasks that need the data (tests, vignettes).
> * We avoid to ship the data with the package, and somehow set it up such that the data is retrieved, preprocessed and installed when needed (or when the package is installed?). This way we can license the package exclusively as Apache-2.0, as the data will be downloaded by the user? If we go for a similar approach, we might actually get rid of CRiSpData as a package, and include the logic of retrieving/installing the example datasets as part of CRiSp/rcoins?
> * Similar to the previous point, we split CRiSpData into a software component hosted on GitHub and licensed  Apache-2.0 and pre-processed data components hosted on Zenodo or figshare with the correct licenses.  CRiSpData then retrieves and setup the datasets from Zenodo when data is required.

@cforgaci:
> Can we store the data and the generating script on 4TU.ResearchData with ODbL 1.0 (available there) and get the data from there as part of the tests/examples/vignettes by the user? In this case the CRiSpData package would not be needed. I think this is similar to one or parts of the solutions you give.

Continuing on my last point: we can retreive the data using the URL from 4TU.RD (or alternative data repository) as part of a wrapper function `get_example_data()` which could return a list with `bucharest_osm` and `bucharest_dem`. We can still keep `bucharest_delineation` (or the renamed version of it as proposed in #137) as part of the package.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
This creates a new module `exampledata` with a couple of functions that retrieve the example OSM and DEM data from 4TU.ResearchData instead of the CRiSpData package. The dataset was published at https://data.4tu.nl/datasets/f5d5e118-b5bd-4dfb-987f-fe10d1b9b386 and the download links for the individual files in the `exampledata` module should work.

## Related Issues

- Related Issue #185 
- Closes #185 

## Added/updated tests?

- [ ] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


vignettes/getting-osm-data.Rmd
```diff
@@ -15,21 +15,18 @@ knitr::opts_chunk$set(
...
-  # nolint start
-  message("Package 'CRiSpData' is required for this vignette but not installed. Install it from GitHub using pak::pak('CityRiverSpaces/CRiSpData')")
-  # nolint end
-  knitr::knit_exit()
-} else {
-  library(CRiSpData)
-}
+
+bucharest_osm <- CRiSp::get_osm_example_data()
+bucharest_dem <- CRiSp::get_dem_example_data()
```
fnattino, line 20: These are already included in lines 27-28, right?
```suggestion
```

tests/testthat/test-corridor.R
```diff
@@ -1,7 +1,7 @@
-test_that("proper parameters must be provided depending on selected method", {
-  skip_if_not_installed("CRiSpData")
+bucharest_osm <- get_osm_example_data()
```
fnattino, line 1: Should we put this and the analogous:
```
bucharest_dem <- get_dem_example_data()
```
In an test helper function (e.g. in `/tests/testthat/helper-example-data.R` ) ? 
This way the datasets should become available to all tests without re-downloading them in all modules.

tests/testthat/test-osmdata.R
```diff
@@ -3,6 +3,8 @@
 # cache folder only used for testing purposes. This is achieved via the
 # [`temp_cache_dir()`] helper function, which should be called in each test.
 
+bb <- get_osm_bb("Bucharest")
+
```
fnattino, line 7: I would move this inside the only function that uses `bb` - it seems to be only one, and so we would also skip it if running in CI.

fnattino (changes requested):
Thanks a lot @cforgaci, nice and much simpler. I only have a few suggestions, the most relevant is probably to move the calls to `get_osm_example_data()`/`get_dem_example_data()` within an helper module so that we do not re-download the example datasets multiple times.

cforgaci (commented):
@fnattino, thanks for the review. I processed most of your suggestions and left a couple of responses.

vignettes/getting-osm-data.Rmd
fnattino: Oh right, thanks for clarifying!

fnattino (approved):
Looks great @cforgaci - good to merge for me!

PR #194, opened 2025-05-12
Return empty sf object if no railways are found

fnattino:
The following error is raised if we try to delineate a region without railways:
```
Error in UseMethod("select") : 
  no applicable method for 'select' applied to an object of class "NULL"
Calls: retrieve_data ... get_network -> get_osm_railways -> <Anonymous> -> <Anonymous>
```

This fix allows `get_osm_railways` to return an empty `sf` object if no railways are found, so that the workflow can continue with the spatial network being based on streets only.


cforgaci (approved):
👍

PR #195, opened 2025-05-20
Improvements for search for city boundaries

### LINKED ISSUE #67: River and city names in OSM queries

fnattino:
* Adding partial matches for names and river - useful to retrieve some input data for the CRiSp dataset. Closes #67 , hope this is good enough even if we don't handle misspellings? 
* Search for city boundary in both OSM polygons and multipolygons (some cities, e.g. "Belfast", would be missed if only multipolygons are considered).
* A message that was printed out every time city boundaries are found is removed to reduce verbosity

### DIFF
--- a/R/osmdata.R
+++ b/R/osmdata.R
@@ -173,7 +173,7 @@ get_osm_city_boundary <- function(bb, city_name, crs = NULL, multiple = FALSE,
   # Define a helper function to fetch the city boundary
   fetch_boundary <- function(key, value) {
     osmdata_sf <- osmdata_as_sf(key, value, bb, force_download = force_download)
-    osmdata_sf$osm_multipolygons |>
+    dplyr::bind_rows(osmdata_sf$osm_polygons, osmdata_sf$osm_multipolygons) |>
       # filter using any of the "name" columns (matching different languages)
       match_osm_name(city_name_clean) |>
       sf::st_geometry()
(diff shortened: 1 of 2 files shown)


fnattino:
Similarly to https://github.com/CityRiverSpaces/rcoins/pull/64 data seems not to be accessible from the 4TU data repo - maybe because of too many calls or are there technical issues at the moment?

fnattino:
> Similarly to [CityRiverSpaces/rcoins#64](https://github.com/CityRiverSpaces/rcoins/pull/64) data seems not to be accessible from the 4TU data repo - maybe because of too many calls or are there technical issues at the moment?

It seems that increasing the timeout seems to fix the issue, will open a new PR to test this

R/osmdata.R
```diff
@@ -183,9 +183,7 @@ get_osm_city_boundary <- function(bb, city_name, crs = NULL, multiple = FALSE,
   city_boundary <- tryCatch(fetch_boundary("boundary", "administrative"),
                             error = function(e) NULL)
 
-  if (!is.null(city_boundary)) {
-    message("City boundary found with 'boundary:administrative' tag.")
-  } else {
+  if (is.null(city_boundary)) {
     stop("No city boundary found. The city name may be incorrect.")
   }
```
cforgaci, line 188: Yes, simpler.

cforgaci (approved):
👍

PR #196, opened 2025-05-23
Increase timeout for example data retrieval

fnattino:
Some of the checks are failing due to the `download.file` function timing out during vignette preparation or tests, this PR tries to solve the issue.

### DIFF
--- a/R/exampledata.R
+++ b/R/exampledata.R
@@ -16,7 +16,10 @@ get_osm_example_data <- function() {
   url_osm <- "https://data.4tu.nl/file/f5d5e118-b5bd-4dfb-987f-fe10d1b9b386/f519315e-b92d-4815-b924-3175bd2a7a61"
   # nolint end
   temp_file <- tempfile(fileext = ".gpkg")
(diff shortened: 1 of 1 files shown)


R/exampledata.R
```diff
@@ -16,7 +16,10 @@ get_osm_example_data <- function() {
   url_osm <- "https://data.4tu.nl/file/f5d5e118-b5bd-4dfb-987f-fe10d1b9b386/f519315e-b92d-4815-b924-3175bd2a7a61"
   # nolint end
   temp_file <- tempfile(fileext = ".gpkg")
-  download.file(url_osm, destfile = temp_file, mode = "wb")
+  # temporarily increate timeout, reset value on exit
+  op <- options(timeout = 300)
+  on.exit(op)
```
cforgaci, line 21: Is `on.exit()` meant to reset the timeout to default here? Shouldn't then this be something like:

```suggestion
  old_timeout <- options(timeout = getOption("timeout"))
  options(timeout = 300)
  on.exit(options(timeout = old_timeout))
```

Issue #134, opened 2025-03-28
Generalising delineation into a separate package


cforgaci:
A recent presentation of CRiSp led to the idea by @fnattino that its delineation functionality could be generalised and extracted into a separate package. Given a any spatial network and any polygon input, such a package should output a polygon which encloses the input polygon and follows the spatial network. In CRiSp, this is equivalent to the street-and-rail network and the DEM-based initial river corridor, respectively.

An exmaple of an input can also be a line in a street network, such as a target street and a custom area of interest around it.

This is low priority now, but I am leaving a note so that we can explore it later.

fnattino:
Closing it for now, not to be tackled/investigated until needed.

PR #201, opened 2025-06-04
Allow for valley to be included in delineation output

### LINKED ISSUE #199: Returning the valley as part of delineation
Besides its usefulness for explaining delineations in plots, the valley boundary may also be useful for some use cases. I think we should return it by default along with corridor, segment and riverspace.

cforgaci:

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
This allows the user to choose whether to include the valley in the final output or not.

## Related Issues

- Related Issue #199 
- Closes #199

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


fnattino:
I like the idea of returning the valley! I have two thoughts on this:
* in the approach implemented here you actually end up computing the valley twice (once within `delineate` and once within `delineate_corridor`) - for both efficiency and code repetition I think we should try to drop one of the two valley estimations. If we want to be consistent to the way in which the other delineations take place, we should probably drop the one inside `delineate_corridor`, and pass the valley as input to it as initial corridor boundary. This would also bring us in the direction of addressing #134, since `delineate_corridor` will become able to work on arbitrary shapes. On the other hand, it will become more difficult to use `delineate_corridor` as stand-alone since one will have to generate the initial corridor as a pre-requisite. I am not yet sure what it the best approach.
* We have already quite a lot of input arguments for `delineate` - what about returning the valley every time it is calculated (i.e. we do the corridor delineation with "valley" as initial method)?

cforgaci:
@fnattino, indeed the valley is computed twice.

> * in the approach implemented here you actually end up computing the valley twice (once within `delineate` and once within `delineate_corridor`) - for both efficiency and code repetition I think we should try to drop one of the two valley estimations. If we want to be consistent to the way in which the other delineations take place, we should probably drop the one inside `delineate_corridor`, and pass the valley as input to it as initial corridor boundary. This would also bring us in the direction of addressing [Generalising delineation into a separate package #134](https://github.com/CityRiverSpaces/CRiSp/issues/134), since `delineate_corridor` will become able to work on arbitrary shapes. On the other hand, it will become more difficult to use `delineate_corridor` as stand-alone since one will have to generate the initial corridor as a pre-requisite. I am not yet sure what it the best approach.

I think it makes sense to remove it from `delineate_corridor` and provide the valley as input. Aren't we doing something similar with `delineate_segments` using the corridor resulting from `delineate_corridor` as input? Also, one can run `delineate` with the defaults to get the same result as `delineate_corridor`, right? I am not entirely sure either.

> * We have already quite a lot of input arguments for `delineate` - what about returning the valley every time it is calculated (i.e. we do the corridor delineation with "valley" as initial method)?

That's fine too. I was just thinking that each of the four delineations should have an explicit parameter, but, indeed, in this case we already set it implicitly with the `initial_method` parameter.

I will be traveling and do not have much time to work on this until next Wednesday. Any chance you can make this change and check if anything else needs to be done before the release of rcrisp? From what I can see besides this PR only the date needs to be added to the changelog. See also #125

fnattino:
Hi @cforgaci, the work on the dataset took longer than expected - I will try to work on this on Tuesday, I'll keep you posted on this.

fnattino:
Please,  go ahead! 🚀

PR #204, opened 2025-06-04
Fix docs in network module

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
This fixes a couple of typos introduced in the documentation of the network module with #189 

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: no need for these doc fixes

## Added entry in changelog?

- [ ] Yes


fnattino (approved):
Thanks for catching the typo! For the multiple arguments on a line, [it seems to be allowed](https://r-pkgs.org/man.html#multiple-arguments), but I am happy to leave it as you have modified - just good to know for other cases :)

cforgaci:
> For the multiple arguments on a line, [it seems to be allowed](https://r-pkgs.org/man.html#multiple-arguments), but I am happy to leave it as you have modified - just good to know for other cases :)

You're right, I changed this back to both params on one line. Initially I thought that was the issue, but it was the typo instead. Assuming that this is still approved, I am merging this PR.

PR #208, opened 2025-06-20
Move valley delineation outside corridor delineation

fnattino:
As discussed in #201 , I am moving the valley delineation outside the corridor delineation.

I gave some thoughts on how to best tackle the case in which we want to use the "buffer" initialization method for the corridor. I came up with this approach which essentially leaves the buffer-based method within the corridor delineation, but allows for arbitrary geometries to be passed in as input.

Do you think this could make sense @cforgaci ?


cforgaci (approved):
Nice! I also like the way you collect the delineations without intermediate objects.

PR #209, opened 2025-06-23
Fix restoring timeout

fnattino:
Same as in rcoins, where the issue was pointed out in the CRAN submission review: https://github.com/CityRiverSpaces/rcoins/pull/73


cforgaci (approved):
👍

PR #218, opened 2025-06-25
Move default cache dir 

### LINKED ISSUE #216: Caching on CRAN
Apparently caching as we are doing it violates CRAN policy. @fnattino, any thoughts on this?

```
Also, this by default populates

   ~/.cache/rcrisp

from

./rcrisp/R/cache.R:  default_dir <- file.path(Sys.getenv("HOME", "."),
".cache", "rcrisp")

This violates the CRAN Policy which says

   Packages should not write in the user’s home filespace (including
   clipboards), nor anywhere else on the file system apart from the R
   session’s temporary directory (or during installation in the location
   pointed to by TMPDIR: and such usage should be cleaned up). Installing
   into the system’s R installation (e.g., scripts to its bin directory)
   is not allowed.

   Limited exceptions may be allowed in interactive sessions if the
   package obtains confirmation from the user.

   For R version 4.0 or later (hence a version dependency is required or
   only conditional use is possible), packages may store user-specific
   data, configuration and cache files in their respective user
   directories obtained from tools::R_user_dir(), provided that by
   default sizes are kept as small as possible and the contents are
   actively managed (including removing outdated material).

so please use tools::R_user_dir() as appropriate.
```

fnattino:
Closes #216  by moving the cache dir to a suitable location, and by adding a check whenever rcrisp is loaded.


cforgaci (approved):
👍

PR #219, opened 2025-06-25
Improve tests

### LINKED ISSUE #217: Overall runtime too high during CRAN's Win checks
From the log:
```
Flavor: r-devel-windows-x86_64
Check: Overall checktime, Result: NOTE
  Overall checktime 13 min > 10 min
```

As seen in this excerpt from the full log, most of the time is spent on tests:
```
* checking examples ... [15s] OK
* checking for unstated dependencies in 'tests' ... OK
* checking tests ... [521s] OK
  Running 'testthat.R' [521s]
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes ... OK
* checking re-building of vignette outputs ... [166s] OK
* checking PDF version of manual ... [23s] OK
* checking HTML version of manual ... [11s] OK
```

We should have a look at which tests can be skipped or made faster to overcome this issue.

fnattino:
Closes #217 

Improve on some tests, specifically:
- for `test-delineate.R`, I have mocked the calls to `get_osmdata` and `get_dem`, so that they can be easily run on the CI. These now represents something closer to integration tests, rather than unit tests.
- for `test-osmdata.R`, I have also introduced a lot of mocking, so tests are faster and do not make actual queries to OSM servers. I have thus removed the `skip_on_ci()` constratints, which will give us a big boost in test coverage as resulting from pkgcheck.

### DIFF
--- a/tests/testthat/test-osmdata.R
+++ b/tests/testthat/test-osmdata.R
@@ -224,7 +224,7 @@ test_that("River retrieval raise error if river is not found in the bb", {
   )
 })
 
-test_that("River lines and surface and properly setup", {
+test_that("River lines and surface are properly set up", {
   with_mocked_bindings(
     osmdata_as_sf = function(...) {
       list(osm_lines = mock_river_lines, osm_polygons = mock_river_polygons)


cforgaci:
@fnattino, I noticed this is ready for review but not assigned to me yet. Let me know when I can have a look.

PR #228, opened 2025-07-02
Add DOI

cforgaci:
I am adding here the DOI information from Zenodo. I also updated the way the title is displayed in CITATION.cff, which is also used for the title on Zenodo — to be updated in the next release.


fnattino (changes requested):
Great, thanks! But maybe let's use the DOI referring to all versions (10.5281/zenodo.15793526)?

fnattino:
One last thought @cforgaci - with the update of the DESCRIPTION file, should the codemeta.json file also be updated?

cforgaci:
> One last thought @cforgaci - with the update of the DESCRIPTION file, should the codemeta.json file also be updated?

I updated it. For some reason `codemetar::write_codemeta()` would not add the DOI link, so I added it manually.

fnattino (approved):
Hope it is still valid! I tried to check with the [codemeta generator](https://codemeta.github.io/codemeta-generator/). It is supposed to be able to validate codemeta files as well, but it raises many errors for fields that I believe are correct (licence, dependencies, etc.) so I don't trust it too much..

PR #235, opened 2025-07-03
Fix namespace load NOTE

cforgaci:
This resolves the NOTE on namespace load. I repalced `.onLoad()` with `.onAttach()` so that `check_cache()` is only triggered when the user executes `library(rcrisp)`. Also, the warning used in `check_cache()` is now replaced by a startup message.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: rcrisp
 Title: Automate the Delineation of Urban River Spaces
-Version: 0.1.3
+Version: 0.1.4
 Authors@R: c(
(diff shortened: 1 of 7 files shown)


fnattino (approved):
Thanks a lot, I think even the simpler:

```
.onAttach <- function(libname, pkgname) {
  check_cache()
} 
```

would have worked, but looks good to me!

cforgaci:
> Thanks a lot, I think even the simpler:
> 
> ```
> .onAttach <- function(libname, pkgname) {
>   check_cache()
> } 
> ```
> 
> would have worked, but looks good to me!

Of course, replacing `.onLoad()` with `.onAttach()` already fixes the issue 🤦‍♂️

PR #253, opened 2025-07-08
Standards in category G2

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
The G2 category turns out to be a rather big task, so let's review first the changes I made and after merging I will create more granular PRs one level lower, for each standard like G2.0.

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes

### DIFF
--- a/tests/testthat/test-corridor.R
+++ b/tests/testthat/test-corridor.R
@@ -119,7 +119,7 @@ test_that("River banks for a simple river gives two regions", {
   river <- sf::st_sfc(sf::st_linestring(cbind(c(-2, 0, 0), c(0, 0, -2))))
   regions <- get_river_banks(river, width = 1)
   expect_equal(length(regions), 2)
-  expect_true(all(sf::st_geometry_type(regions) == "POLYGON"))
+  expect_true(inherits(regions, "sfc_POLYGON"))
 })
 
 test_that("River banks for a more complex river still gives two regions", {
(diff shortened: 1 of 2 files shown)


cforgaci (commented):
To comply with the G2 standards and avoid repetition, we will need some sort of `validate_input()` function that checks lenghts, types, univariate input. I see that the `checkmate` package can be used for that.

tests/testthat/test-utils.R
```diff
@@ -115,6 +115,8 @@ test_that("buffering a bbox does not change its CRS", {
   expect_equal(crs_actual, crs_expected)
 })
 
+#' @srrstatsTODO {G2.4, G2.4c} Explicit conversion to character with
+#'   `as.character()` used to test the output type of `buffer()`.
```
cforgaci, line 119: I am not sure if test code like this falls within the scope of G2.4. Maybe all test code should fall under G5.

fnattino: Not sure about tests, but we should check all uses oof`sf::st_geometry_type()`, which returns a factor - so it should involve G2.4e:
```
G2.4e explicit conversion from factor via `as...()` functions
```
Here, for instance we use `as.character()`

fnattino (commented):
Thanks! Just a small comment on the use of `sf::st_geometry_type`, which returns a factor.

R/delineate.R
```diff
@@ -52,6 +56,10 @@ delineate <- function(
 
   delineations <- list()
 
+  if (length(city_name) > 1 || length(river_name) > 1) {
+    stop("Only one city and one river can be delineated at a time.")
+  }
+
```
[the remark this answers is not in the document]
cforgaci: Yes, that is correct! I would merge #265 first and then update this.

tests/testthat/test-utils.R
cforgaci: Updated with the last one.

PR #256, opened 2025-07-08
Standards in category SP3

cforgaci:
None of the standards in this category seem to be applicable.

### DIFF
--- a/R/srr-stats-standards.R
+++ b/R/srr-stats-standards.R
@@ -117,7 +117,6 @@ NULL
 #'   spatial neighbours.
 #' @srrstatsNA {SP3.2} The package does not rely on sampling from input data.
 #' @srrstatsNA {SP3.3} The package does not employ regression.
-#' @srrstatsNA {SP3.4} The package does not use any kind of clustering.
 #' @srrstatsNA {SP3.5, SP3.6} The package does not implement any kind of
 #'   machine learning.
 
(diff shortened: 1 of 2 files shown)


R/srr-stats-standards.R
```diff
@@ -122,5 +113,13 @@ NULL
 #'   and no performance claims are made.
 #' @srrstatsNA {G1.6} As there are no alternative implementations, no
 #'   performance claims are made in this package.
+#' @srrstatsNA {SP3.0, SP3.0a, SP3.0b, SP3.1} The package does not consider
+#'   spatial neighbours.
+#' @srrstatsNA {SP3.2} The package does not rely on sampling from input data.
+#' @srrstatsNA {SP3.3} The package does not employ regression.
+#' @srrstatsNA {SP3.4} The package does not use any kind of clustering.
```
fnattino, line 120: We use dbscan as a clustering method in order to figure out "independent" river crossings:

https://github.com/CityRiverSpaces/rcrisp/blob/c0fde0578b59d70bebcc72ffe6025254063cd59c/R/segments.R#L120

 In a certain way the clustering is "spatial" (it's also the "S" of DBSCAN), so I think we can say we fulfil this standard.

fnattino (changes requested):
Just one comment on the use of clustering.

fnattino (approved):

PR #268, opened 2025-07-16
Get rid of all comparisons including `st_geometry_type`

fnattino:
To address https://github.com/CityRiverSpaces/rcrisp/pull/253#discussion_r2207364830


cforgaci (approved):
👍

PR #270, opened 2025-07-17
Replace sapply with the safer vapply

cforgaci:


R/segments.R
```diff
@@ -168,8 +168,9 @@ select_nonintersecting_lines <- function(lines, corridor) {
   } else {
     # Identify the line with maximum number of intersections
     intersecting_lines <- intersections[["origins"]]
-    num_intersections <- sapply(seq_len(length(lines)),
-                                \(x) sum(unlist(intersecting_lines) == x))
+    num_intersections <- vapply(seq_len(length(lines)),
+                                \(x) sum(unlist(intersecting_lines) == x),
+                                numeric(1))
```
fnattino, line 173: Shouldn't this return an integer?
```suggestion
                                integer(1))
```

fnattino (changes requested):
Thanks, only one comment - correct me if I am wrong!

cforgaci:
> Should we add an entry to the changelog?

Thanks for reminding me. I updated the changelog.

fnattino (approved):
👍

PR #275, opened 2025-07-18
State compliance with G2.2

cforgaci:


fnattino (commented):
Sounds good, should this be moved to the lower section as a `@srrrstats` ?

fnattino (approved):

PR #276, opened 2025-07-18
Match explicit conversions with G2.4 standards

cforgaci:
I identified one possible case for each of the following:

> **Uni-variate (Vector) Input**
> It is important to note for univariate data that single values in R are vectors with a length of one, and that 1 is of exactly the same data type as 1:n. Given this, inputs expected to be univariate should:
> - G2.4 Provide appropriate mechanisms to convert between different data types, potentially including:
>   - G2.4a explicit conversion to integer via [as.integer()](https://rdrr.io/r/base/integer.html)
>   - G2.4b explicit conversion to continuous via [as.numeric()](https://rdrr.io/r/base/numeric.html)
>   - G2.4c explicit conversion to character via [as.character()](https://rdrr.io/r/base/character.html) (and not paste or paste0)

Now, I say "possible", because I am not sure if the cases I identified are indeed what is targeted with these input-related standards. We use casting for testing purposes, within an internal function and to generate a cache file name. It might even be that G2.4 and its sub-standards are not applicable at all.


fnattino (approved):
Sounds good!

PR #269, opened 2025-07-17
Update input assertions and descriptions

cforgaci:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
Update input assertions and descriptions throughout the package to comply to various standards in standard category G2.

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?
_We encourage you to keep the code coverage percentage at 75% and above._

- [x] Yes
- [ ] No, and this is why: _please replace this line with details on why tests
      have not been included_
- [ ] I need help with writing tests

## Added entry in changelog?
_For user-facing changes, add a line describing the changes in [NEWS.md](/NEWS.md)_

- [ ] Yes

PR #278, opened 2025-07-19
Comply with standard G2.16

cforgaci:


R/segments.R
```diff
@@ -23,13 +23,18 @@
...
                                angle_threshold = 100) {
   # Check input
   checkmate::assert_class(corridor, "sfc_POLYGON")
   checkmate::assert_class(network, "sfnetwork")
   checkmate::assert_true(inherits(river, c("sf", "sfc")))
-  checkmate::assert_numeric(angle_threshold, lower = 90, upper = 180)
+  checkmate::assert_numeric(angle_threshold,
+                            lower = 90,
+                            upper = 180,
+                            any.missing = FALSE)
```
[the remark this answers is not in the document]
cforgaci: `lower` and `upper` already rule out `-Inf` and `Inf` so that additional argument is not useful.

R/segments.R
```diff
@@ -23,13 +23,18 @@
...
+#'   (NaN, Inf, -Inf) and errors when encountering such values.
 delineate_segments <- function(corridor, network, river,
                                angle_threshold = 100) {
   # Check input
   checkmate::assert_class(corridor, "sfc_POLYGON")
   checkmate::assert_class(network, "sfnetwork")
   checkmate::assert_true(inherits(river, c("sf", "sfc")))
-  checkmate::assert_numeric(angle_threshold, lower = 90, upper = 180)
+  checkmate::assert_numeric(angle_threshold,
+                            lower = 90,
```
fnattino, line 35: I think this formally should be in the [0;180) range.
```suggestion
                            lower = 0,
```

R/osmdata.R
```diff
@@ -333,11 +333,16 @@ get_osm_river <- function(bb, river_name, crs = NULL, force_download = FALSE) {
...
+#'   (NaN, Inf, -Inf) and errors when encountering such values.
 get_osm_streets <- function(aoi, crs = NULL, highway_values = NULL,
                             force_download = FALSE) {
   # Check input
   checkmate::assert_true(inherits(aoi, c("sf", "sfc", "bbox")))
-  checkmate::assert_numeric(crs, null.ok = TRUE)
+  checkmate::assert_numeric(crs,
+                            null.ok = TRUE,
+                            any.missing = FALSE,
+                            finite = TRUE)
```
fnattino, line 345: This check feels weird - and I think will be superseded by #298 and #269 
```suggestion
  checkmate::assert_numeric(crs, null.ok = TRUE)
```

fnattino (changes requested):

PR #288, opened 2025-07-21
State compliance with standards SP2.4

cforgaci:


fnattino (approved):
Looks good, just wondering whether #309 would also add to this.

PR #293, opened 2025-07-21
State compliance with SP2.8 and SP2.9

cforgaci:


fnattino (approved):
Looks good, even though I guess it would be nice to have more examples of this ..

cforgaci:
> Looks good, even though I guess it would be nice to have more examples of this ..

We are doing this with #311 as suggested by @fnattino. Merging this PR.

PR #298, opened 2025-07-21
Apply standards SP6.1

cforgaci:


fnattino (commented):
I have tried to run `delineate_corridor` on a dataset with geographic CRS and things go wrong in the function that calculates weights. Maybe we should a check that the CRS is projected (not geographic) and raise an error otherwise?

cforgaci:
> I have tried to run `delineate_corridor` on a dataset with geographic CRS and things go wrong in the function that calculates weights. Maybe we should a check that the CRS is projected (not geographic) and raise an error otherwise? What do you think?

Thanks. That seems like a reasonable solution. Delineating on geographic CRS does not make much sense anyway, right? I will make that change later and ping you again here when I am done.

cforgaci:
@fnattino after #311 is merged, I will update this PR accordingly. `as_crs(x, allow_geographic = FALSE)` should by default error when geographic CRS is used.

Now, considering that all `delineate_*()` functions can be called independently by the user, shouldn't the CRS be checked there instead of in `delineate()`? We could first check that the input objects do not raise the error `"The input CRS is geographic (lat/lon), please provide a projected CRS."` with `as_crs(x, allow_geographic = FALSE)`. Would that work?

fnattino (approved):
Looks good @cforgaci, thanks!

PR #307, opened 2025-07-22
Cache example data files

### LINKED ISSUE #266: Cache example data after retreival to be used across tests and vignettes

fnattino:
Closes #266


R/exampledata.R
```diff
@@ -1,9 +1,11 @@
 #' Get example OSM data
 #'
-#' This function retrieves example OpenStreetMap (OSM) data from a persistent
-#' URL on the 4TU.ResearchData data repository, and it can be used in examples
+#' This function retrieves example OpenStreetMap (OSM) data from a
```
cforgaci, line 3:
```suggestion
#' This function retrieves example OpenStreetMap (OSM) data from the
```

cforgaci (approved):
Wow, let's hope this fixes the issue. Thanks!

fnattino:
@cforgaci this implementation now uses the Zenodo data repository. Can you have a look when you have time? 

All checks seem to pass for now, we can try to merge it and see whether this approach is more stable. 

The only uncertainty I have for the current setup is related to versioning of the example dataset. Right I am building the download URLs from the record ID, but I guess this will change when a new version of the dataset is published? And since we rely on 4TU.RD to release the DOI, we do not have an "overarching" DOI that we can use to reference all versions?

But maybe we can figure this out later, when we release a new version of the dataset. If we see that Zenodo works better, the easiest thing would be to use Zenodo as "main" data repository, and potentially add links to 4TU.RD.

PR #308, opened 2025-07-23
Add fix to UTM zone finder and tests for polar regions

fnattino:


tests/testthat/test-utils.R
```diff
@@ -44,6 +44,23 @@ test_that("correct UTM zone is returend in the northern hemisphere", {
   expect_equal(utm_epsg, 32627)
 })
 
+#' @srrstats {SP6.2} The software works mostly in projected coordinates. By
+#'   default we work with UTM zones, which are defined only outside polar
+#'   regions. He we make sure that errors are raised if polar regions are
```
cforgaci, line 49:
```suggestion
#'   regions. We make sure that errors are raised if polar regions are
```

tests/testthat/test-delineate.R
```diff
@@ -82,8 +82,6 @@ test_that("Only one city and one river can be delineated at a time", {
                "Assertion on 'river_name' failed: Must have length 1")
 })
 
-#' @srrstats {SP6.2} This test checks that delineation is successfully carried
-#'   out for a river-crossed town close to the North Pole.
```
cforgaci, line 86: If we drop this annotation because the case is not extreme enough, does it still make sense to keep this test?

fnattino: OK, then let's remove it!

PR #310, opened 2025-07-24
Remove message suppression with sf warning

fnattino:
These messages ([this](https://github.com/CityRiverSpaces/rcrisp/pull/265/files#r2207486200) and [this](https://github.com/CityRiverSpaces/rcrisp/pull/265/files#r2207487075))are indeed the informative messages from SF on the less suitability to use geographic CRS here, so we might remove them. The autotest issue will probably get solved since non-geographic CRSs will be stopped earlier.

### DIFF
--- a/R/network.R
+++ b/R/network.R
@@ -81,7 +81,7 @@ flatten_network <- function(network) {
 #' @noRd
 get_crossing_edges <- function(edges) {
   geometry <- sf::st_geometry(edges)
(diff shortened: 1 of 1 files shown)


cforgaci (approved):
Okay! Anyway, I realised that it makes more sense to run autotest only after all standards are applied. Otherwise, it is an endless back-and-forth process.

PR #320, opened 2025-07-26
Update method diagram

cforgaci:
While working on the position paper, I updated the method diagram. This version is:

- more compact, with the layers vertically stacked closer together, so easier to see on the screen;
- more consistent with the four delineations we are describing in the vignette.


fnattino (approved):
Oh nice, looks great!

PR #321, opened 2025-07-29
Fix autotest

cforgaci:
The changes proposed in this PR address the issues raised by `autotest::autotest_pacakge()`. After we merged, `rcrisp` should be ready for rOpenSci's Statistical Software Review.


tests/testthat/test-network.R
```diff
@@ -62,6 +62,7 @@ network_no_crossings <- sfnetworks::sfnetwork(nodes = nodes_no_crossings,
                                               edges = edges_no_crossings,
                                               directed = FALSE, force = TRUE,
                                               node_key = "x")
+sf::st_crs(network_no_crossings) <- 32635
```
cforgaci, line 65: This is needed since `as_crs()` was introduced.

tests/testthat/test-autotest.R
cforgaci, line 1: I tried to make this work, but it's been quite difficult. The test runs and does lead to a `NULL` autotest result if:
- it is run interactively; and
- the installed version of `rcrisp` is used (devtools::install(); library(rcrisp)`), not the source version (`devtools::load_all()`), because autotest needs to see the latest version of the Rd files in the documentation.

Note that:
- After (too) many iterations, I managed to turn off named tests safely with justification, as indicated in the autotest documentation.
- I had to remove rows that result from normal function calls after running `autotest_package()` with `test = TRUE`; those cannot be excluded with `test_data` because they do not have `test_name`s.
- For some reason, some of the removed named tests remain after running `autotest_package()` with `test = TRUE`, so I had to remove those manually.
- Because of the manual steps, `autotest::expect_autotest_testdata()` cannot be run as indicated in the [documentation](https://docs.ropensci.org/autotest/articles/autotest-control.html#autotest-ing-your-package)

PR #324, opened 2025-08-19
Release 0.2.0 and CRAN fix

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [x] Documentation Update

## Description
This bumps the version to 0.2.0 and fixes the current issue on CRAN by updating `get_osm_example_data()` and `get_dem_example_data()` to fail gracefully in examples, vignettes and tests if internet resource is not available.

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: no changes in functionality

## Added entry in changelog?

- [x] Yes

### DIFF
--- a/cran-comments.md
+++ b/cran-comments.md
@@ -4,10 +4,10 @@
 
 ## Resubmission
 
-* To fix the current problems on CRAN: `get_osm_example_data()` and
-`get_dem_example_data()` were updated to fail gracefully in examples, vignettes
-and tests, that is, return a message and NULL, if internet resource is not
-available.
+* This is a resubmission to solve problems identified by package checks on CRAN.
+Specifically, we have updated `get_osm_example_data()` and
+`get_dem_example_data()` to fail gracefully in examples, vignettes and tests,
+that is, return a message and NULL, if the internet resource is not available.
 
 * In addition, this version includes extensive documentation updates, adds input
 assertions throughout the package, removes an unused function, and replaces


fnattino:
Thanks a lot @cforgaci , looks great! The solution for the vignettes that use the example dataset looks nice. Unfortunately, I still see that one action is failing on vignette 5, where data is acessed directly from OSM data (for which we have no control). One potential solution would be to wrap the call to delineate withing a `tryCatch` block, to gracefully quit upon all HTTP related errors. Another solution could be something similar to what I have done for rcoins in https://github.com/CityRiverSpaces/rcoins/pull/83, i.e. freezing the vignette and checking in the precomputed article..

cforgaci:
> Unfortunately, I still see that one action is failing on vignette 5, where data is acessed directly from OSM data (for which we have no control). One potential solution would be to wrap the call to delineate withing a `tryCatch` block, to gracefully quit upon all HTTP related errors.

I isolated the call to `delineate()` into a separate code chunk and disabled evaluation, as the output happens to be `bucharest_dambovita` which is already available as package data. I guess this should not raise any further issues.

> Another solution could be something similar to what I have done for rcoins in [CityRiverSpaces/rcoins#83](https://github.com/CityRiverSpaces/rcoins/pull/83), i.e. freezing the vignette and checking in the precomputed article..

If what I did is good enough to fix the issue on CRAN, I would leave it like this and implement the solution you used in `rcoins` later, if needed. Do you agree?

PR #325, opened 2025-08-21
Remove periods from titles in function docs

cforgaci:


fnattino (approved):
Nice that you caught this, thanks!

fnattino:
Hey, but the errors I see in the action do not seem related to remote access, rather to some input argument (`doc`?) missing within some internal osmdata functions?

cforgaci:
> Hey, but the errors I see in the action do not seem related to remote access, rather to some input argument (`doc`?) missing within some internal osmdata functions?

True, I realised after I wrote the message.. as I only made minor ortographic changes in the documentation, this is probably an osmdata retrieval issue which should not arise anyway if the OSM vignette is updated, right?

PR #326, opened 2025-08-21
Update vignette

cforgaci:
@fnattino, I remember discussing with you that in this vignette it makes sense to run the actual OSM requests, but to avoid further issues related to HTTP requests on CRAN, I disabled those code chunks.

Now, I notice an issue with the example data: the street network includes the full extent whereas the actual function calls would only get the streets and railways within the given buffer. How should we deal with that?

### DIFF
--- a/vignettes/vig_02-getting-osm-data.Rmd
+++ b/vignettes/vig_02-getting-osm-data.Rmd
@@ -7,13 +7,26 @@ vignette: >
   %\VignetteEncoding{UTF-8}
 ---
 
-```{r setup, include = FALSE}
+```{r setup, include=FALSE}
 knitr::opts_chunk$set(
   collapse = TRUE,
   comment = "#>",
   message = FALSE,
-  warning = FALSE
+  warning = FALSE,
+  eval = FALSE
 )
+```
+
+```{r data, eval=TRUE, echo=FALSE}
+library(rcrisp)
+
+bucharest_osm <- get_osm_example_data()
+
+data_available <- !is.null(bucharest_osm)
+
+if (!data_available) {
+  cat("NOTE: OSM data could not be downloaded; result will be not be plotted.")
+}
 ```
 
 In this notebook we download OpenStreetMap (OSM) data needed for the delineation of the urban river corridor of River Dâmbovița in Bucharest, Romania. After attaching the packages used in the vignette, we specify the city name, the river name, and the CRS, and we make sure that we provide a buffer around the river used to retrieve OSM data.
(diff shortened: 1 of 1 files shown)


fnattino:
I had completely forgot about this vignette. 

So, options we have are:

1. Modify the dataset, but I don't think this is a good choice, since we use the same dataset for rcoins as well (for which having the full city is nice).

2. Add some more hidden code to the vignette, to build data like the one retrieved by `rcrisp::get_osm_data()` starting from the example dataset. To me, this also does not seem like a good approach - the extra code in the vignette will be difficult to maintain, and the point of the vignette of illustrating the use of the code will be partly deceived (hidden code, "fake" output).

3. Leave the vignette with the actual content that we want to show, but run the vignette only locally and freeze its output (same as what I did in rcoins). This solves the problem, with the pro that the output of the vignette will be properly displayed online, and the cons that one has to manually run vignette generation every once in a while (code might become broken).

4. Leave the vignette with the actual content that we want to show, but disable running of the vignette (set `eval=False`). This also solves the problem at the root, with the pro that we don't have to take care of the vignette anymore and there are no manual steps involved, and cons that the vignette might become broken and the output of the vignette will not be displayed online.

I would go for either 3. or 4. . The latter is also much faster to implement, so we might also go for it and leave improvement (vignette freezing?) for later..

cforgaci:
> I would go for either 3. or 4. . The latter is much faster to implement, so we might go for it and leave improvement (vignette freezing?) for later, if in a hurry.

@fnattino thanks! I will go for 4, to make sure we submit to CRAN today. Will let you know here when I am done.

vignettes/vig_02-getting-osm-data.Rmd
```diff
@@ -37,46 +51,62 @@ bb <- get_osm_bb(city_name)
...
+plot(bucharest_osm$streets$geom, color = "black", add = TRUE)
 plot(bucharest_osm$river_surface, col = "blue", border = "blue", add = TRUE)
 plot(bucharest_osm$river_centerline, col = "blue", add = TRUE)
+plot(bucharest_osm$boundary, border = "red", add = TRUE)
 ```
 
 Individual layers can be written to disk before being read in for delineation.
 
-```{r eval=FALSE}
+```{r, eval=FALSE}
```
cforgaci, line 109: I would still keep eval=FALSE here explicit to make sure it is not switched on later.

cforgaci:
> I will go for 4, to make sure we submit to CRAN today. Will let you know here when I am done.

@fnattino will this be sufficient?

PR #335, opened 2025-09-09
Error when river has no crossings

### LINKED ISSUE #121: Delineation along streams in peripheral locations fails
In peripheral locations, streams tend to be far away from streets or they are only followed by one street. What should we do in those cases? Increase the buffer until corridor boundaries can be found on both sides? Or fall back on using a buffer? Or refuse to delineate such segments? In any case with a warning?

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
If the network does not cross the river at all, delineation will fail with an informative error.

## Related Issues

- Related Issue #121
- Closes #121

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes

### DIFF
--- a/tests/testthat/test-corridor.R
+++ b/tests/testthat/test-corridor.R
@@ -341,3 +341,19 @@ test_that("Errors are raised for wrong input types to corridor delineation", {
   expect_error(delineate_corridor(network_edges, river),
                "Assertion on 'network' failed")
 })
+
+test_that("When river has no crossing, delineation fails with informative
+          error.", {
+            river <- sf::st_sfc(sf::st_linestring(cbind(c(-3, 0.5), c(0, 0))),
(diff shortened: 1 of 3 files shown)


tests/testthat/test-corridor.R
cforgaci, line 1: @fnattino when I try to run this test, R crashes, and I cannot figure out why.

cforgaci: @fnattino the issue was in building the river network. After adding a third node to the river, the test seems to be working now, but I cannot tell why that fixes the issue. Maybe something to tackle outside this PR if the issue occurs again.

PR #348, opened 2025-11-12
Disable code chunk evaluation in network vignette

### LINKED ISSUE #347: Network vignette failing on CRAN
The network vignette is [failing on CRAN](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-gcc/rcrisp-00check.html). We need to submit a fix before 2025-11-25.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
Code chunk evaluation is disabled to avoid errors from failed http requests on CRAN. I also precumputed the plot and added it as an image via markdown.

## Related Issues

- Related Issue #347
- Closes #347

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: no functionality has changed
- [x] I need help with writing tests: however, @fnattino, is there a way to test if the package (incl. vignettes, examples and tests) fails gracefully on failed http requests? I see [an article on rOpenSci](https://books.ropensci.org/http-testing/graceful.html) referring to [devtools::release_questions()](https://devtools.r-lib.org/reference/release.html#details)

## Added entry in changelog?

- [x] Yes


fnattino (approved):
Looks good, thanks. The more I see these issues, the more I think the solution of [pre-computing vignettes](https://blog.r-hub.io/2020/06/03/vignettes/#how-to-include-a-compute-intensive--authentication-dependent-vignette), also linked in the chapter of rOpenSci you linked, is the way to go if you have any dependence on remote data..  

Concerning testing for "graceful" failures: as far as I understand `devtools::release_questions()` can be used to only add reminders that pops up when you run `devtools::release()` - a way that one might use to remind yourself of updating a "frozen" vignette rather than a real check. 

I am not sure about whether there is a straightforward way to test for graceful vignettes/tests/example failures. The easiest way would be to try to run something like `R CMD check` offline - but there seem to be no such option. 

Alternatively, one could work on the code and make sure that all functions fail gracefully when resources are not available. This would probably help in the tests (which BTW, are currently safe for rcrisp), but not in the examples and vignettes: one should embed all the logic of skipping subsequent code blocks in the examples/vignettes themselves (and yet, you would not have a mechanism to effectively check on whether the behaviour is correct)..

cran-comments.md
```diff
@@ -5,8 +5,16 @@
...
+* We also noticed in the rendering of all vignettes a warning originating from
+`rmarkdown`. A recent release of `rmarkdown` (v2.30) addresses this issue and
+thus we expect that the warning will be resolved when that version is used in
+the checks on CRAN.
+
+* One of the CRAN checks also shows a dependency error, stating that `osmdata`
+is not available. However, `osmdata` is on CRAN and we believe this to be a
+temporary issue on CRAN's side. (Note that the `osmdata` check results form the
+failing flavor "r-devel-linux-x86_64-fedora-gcc" seem to be unavailable:
+https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-gcc/osmdata-00check.html)
```
cforgaci: I am not sure if I should add this note. I noticed that all other checks are available expect the system where osmdata is not available for crisp.

fnattino: Uhm, maybe it was not available earleier, but the `osmdata` check for `r-devel-linux-x86_64-fedorac-gcc` seems to be available (and OK) now: https://cran.r-project.org/web/checks/check_results_osmdata.html ? 
I would just point to the failing check: 
```suggestion
temporary issue on CRAN's side (see check for `r-oldrel-macos-x86_64`: https://www.r-project.org/nosvn/R.check/r-oldrel-macos-x86_64/rcrisp-00check.html).
```

fnattino (changes requested):
~~Just noticed we have yet another failing check due to tests not being able to access the  data - even though I thought we have skipped all tests with dependencies on the example data on CRAN. I am going to try to investigate this a bit further.~~  

https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-gcc/rcrisp-00check.html

EDITED: sorry, I had a closer look at the failed check above and I thought it was a different check from the one which we are targeting with this PR. However, I think this PR solves the issue with the vignette, but the CRAN check also reveals that we still have problems with tests, if the example data turns out not to be available.  Considering the amount of problems we are having with the retrieval of example data as part of CI and CRAN regular checks, I would suggest the following: 

* let's create some very small artificial dataset, which we can add to this repository without licensing issue. We could use simple geometric shapes or use generative AI to produce more realistic datasets. Let's use these datasets in all tests that periodically run on CRAN and on CI (and in examples?). Let's skip on both CRAN and CI all tests and examples that use the example data that we have on Zenodo.

* Let's freeze the vignettes by using the pre-computing technique linked above. The drawback is that one needs to remember do periodically update the vignettes (e.g. before each release). We could make a list of all tasks to be carried out before the release and add it to the developer's instructions (do we have these? If not, let's create them :) ).

What do you think @cforgaci ? If you agree, I could take care of these tasks within this week.

cforgaci:
> What do you think @cforgaci ? If you agree, I could take care of these tasks within this week.

@fnattino many thanks for looking into this! I agree with your proposal. I cannot work on this until the CRAN deadline on 25 November, so it would be great if you can take care of it.

fnattino:
There's a failing test - having a look at it now.

fnattino (approved):
Hi @cforgaci  this looks ready to me - let me know if there is anything else I can help with here!

cforgaci:
> > What do you think @cforgaci ? If you agree, I could take care of these tasks within this week.
> 
> @fnattino many thanks for looking into this! I agree with your proposal. I cannot work on this until the CRAN deadline on 25 November, so it would be great if you can take care of it.

Thanks, @fnattino! I will update the date in the changelog, merge and resubmit to CRAN later today.

cforgaci:
@fnattino, FYI, rcrisp 0.3.1 is on CRAN.

PR #349, opened 2025-11-19
Freeze vignettes, make examples and tests free from remote resources

fnattino:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [x] Optimization
- [x] Documentation Update

## Description

- The checks on examples run by `R CMD check` now passes even offline.  
- Vignettes are frozen, so that they can be compiled offline as well. They can be updated by running the `precompile.R` script.
- Tests do not make use of the example data anymore. In some of the tests I have replaced some very simple artificial data, and manually created some slightly more complex artificial data for tests that requires this. Tests can now also be run offline.

## Related Issues

- Builds on #348 

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes


codemeta.json
fnattino, line 1: Cannot figure out why updating the codemeta.json with `codemetar::write_codemeta()` removes the Zenodo DOI link - but it is probably not a big deal if that is not linked here.

R/utils.R
fnattino, line 1: This was the only example that would require remote resources (and would not have the `ifInteractive()` flag)

fnattino:
@cforgaci - this should be it. `R CMD check` passes locally - even if I put my laptop offline, so I really hope this should solve all issues with remote data access on CI/CRAN. 

For the vignettes: I have followed the approach described [here](https://ropensci.org/blog/2019/12/08/precompute-vignettes/), with the `.Rmd` being "static" files generated from the corresponding `.Rmd.orig` files when running `precompile.R` (first remove the `.Rmd` files before regenerating). Precompiling the vignettes also generates all the images, which are saved in `img`. 

NOTE: I have noticed that for some reason all aspect ratii of the figures that I generate locally are one (all figures are square), while the figures that we have in the vignettes online have shapes that more nicely adapt to the figure content. Probably not a big deal, but would you mind trying to run `precompile.R` yourself and see whether you obtain nicer figures? If this is the case, you could check them in here. I have also left the resolution to the default (72 dpi?), we could maybe slightly increase this (`dpi=XXX` in the cell options), but let's pay attention to the limit of 5 MB for the package size in order to make it fit on CRAN ...

vignettes/vig_01-method.Rmd
cforgaci, line 1: This also means that the `@srrstats` tags are only visible in the Rmd.orig files. Maybe good to mention this when leaving an update in the rOpenSci review issue.

codemeta.json
cforgaci: I noticed that too and I usually discard the change in that line after running `codemetar::write_codemeta()`. Can we do that?

vignettes/img/method-diagram.png
cforgaci, line 1: I reduced the size of the methods diagram, because the overall package size exceeded 5MB with the frozen vignette figures.

cforgaci (approved):
Thank you, @fnattino! Regarding the vignette figures, I get the same aspect ratios; I guess this is not a big issue.

codemeta.json
```diff
@@ -4,7 +4,7 @@
   "identifier": "rcrisp",
   "description": "Provides tools to automate the morphological delineation of riverside urban areas based on a method introduced in Forgaci (2018) <doi:10.7480/abe.2018.31>. Delineation entails the identification of corridor boundaries, segmentation of the corridor, and delineation of the river space using two-dimensional spatial information from street network data and digital elevation data in a projected CRS. The resulting delineation can be used to characterise spatial phenomena that can be related to the river as a central element.",
   "name": "rcrisp: Automate the Delineation of Urban River Spaces",
-  "relatedLink": ["https://cityriverspaces.github.io/rcrisp/", "https://CRAN.R-project.org/package=rcrisp", "https://doi.org/10.5281/zenodo.15793526"],
+  "relatedLink": ["https://cityriverspaces.github.io/rcrisp/", "https://CRAN.R-project.org/package=rcrisp"],
```
fnattino:
```suggestion
  "relatedLink": ["https://cityriverspaces.github.io/rcrisp/", "https://CRAN.R-project.org/package=rcrisp", "https://doi.org/10.5281/zenodo.15793526"],
```

codemeta.json
fnattino: Done!

vignettes/vig_01-method.Rmd
fnattino: Exactly - I had a quick look at options to make the cells only visible in the "intermediate" .Rmd  files, but could not find any..

PR #361, opened 2026-02-22
Fix mismatches between SRR annotations and reality

### LINKED ISSUE #354: roxygen + SRR annotation should match actual return structure/classes
As pointed out in the rOpenSci review,

> several documented/SRR “claims about the package” are internally inconsistent with what the code/tests actually do. [...] the package could be brought to a much stronger position with a focused round of revisions that (a) makes declared returns match reality, and (b) treats SRR annotations as audited factual statements that are continuously kept consistent with the code.

More specifically:
  
> * `delineate()` return is wrong/incomplete (SP4.2 / SP4.0 / SP4.1).
>   
>   * Docs currently claim a list of `sf::sfc_POLYGON` objects for corridor/segments/riverspace (`R/delineate.R:41`), but implementation/tests allow conditional `valley` and `sfc_MULTIPOLYGON` (`R/delineate.R:128`, `R/delineate.R:134`, `tests/testthat/test-delineate.R:35`). Please, update roxygen + SRR annotation to match actual return structure/classes (`R/delineate.R:67`).
> * `clear_cache()` documents a return value it does not return.
>   
>   * Docs claim a “list of removed file paths” (`R/cache.R:159`), but the function does not return that (`R/cache.R:168`, `R/cache.R:190`). You should either change the documentation to state that the function returns `NULL`, or change the function to actually return the removed paths.
> * `get_dem()` is not actually case-insensitive, contradicting its SRR claim (G2.3b).
>   
>   * SRR claims case-insensitive handling via `toupper()` (`R/valley.R:54`), but validation happens before uppercasing (`R/valley.R:65`, `R/valley.R:70`). You need to validate after normalisation, or remove the misleading SRR statement and document strict case sensitivity explicitly.
> * Package-level SRR claim for G3.0 conflicts with at least one numeric equality comparison.
>   
>   * SRR claims “no floating point equality comparisons” (`R/srr-stats-standards.R:20`), yet code uses `distances == max(distances)` (`R/corridor.R:213`). You should avoid the equality pattern (or qualify the SRR statement with an explicit rationale and make it precise rather than categorical).

cforgaci:

- [x] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
This fixes the 

## Related Issues

- Related Issue #354 
- Closes #354

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed typos in test statements.
 - Corrected documentation for `reproject()`: the `crs` parameter no longer (incorrectly) lists `logical` as an accepted type. `crs` accepts numeric/integer, character (e.g. "EPSG:4326") or an `sf::crs` object; passing `TRUE`/`FALSE` will fail.
 - Vignette pre-compilation was updated so that srr tags dropped by `knitr::knit()` are reinserted into the vignettes. Affected vignettes were also recompiled.
+- Documentation of return values in `clear_cache()` and `delineate()` were made factually consistent with the code.
 
 ## Removed
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -7,13 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Fixed
 
-<<<<<<< HEAD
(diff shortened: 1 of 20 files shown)


fnattino (approved):
Thanks!

PR #364, opened 2026-02-24
Attach-time fixes

### LINKED ISSUE #359: Attach-time side effects and warning formatting
> * Attach-time side effects and warning formatting: `check_cache()` on attach (`R/zzz.R:2`), directory creation side effects (`R/cache.R:17`), and a warning string that appears to have an unmatched parenthesis (`R/cache.R:214`). Consider gating attach-time checks behind `interactive()`/options and fixing the warning text.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Related Issues

- Related Issue #359 
- Closes #359

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: is a test needed here?

## Added entry in changelog?

- [x] Yes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,10 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Missing parenthesis was added in warning text in `check_cache()`.
 - Vignette pre-compilation was updated so that srr tags dropped by `knitr::knit()` are reinserted into the vignettes. Affected vignettes were also recompiled.
+- Fixed typos in test statements.
+- Corrected documentation for `reproject()`: the `crs` parameter no longer (incorrectly) lists `logical` as an accepted type. `crs` accepts numeric/integer, character (e.g. "EPSG:4326") or an `sf::crs` object; passing `TRUE`/`FALSE` will fail.
+- Vignette pre-compilation was updated so that srr tags dropped by `knitr::knit()` are reinserted into the vignettes. Affected vignettes were also recompiled.
 
 ## Changed
 
 - Attach-time `check_cache()` was moved behind `interactive()`.
+
+## Removed
+
+- TODOs were moved from released code to GitHub issues.
 
 # rcrisp 0.3.1 - 2025-11-24
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# [Unreleased]
+
+## Changed
+
+- Attach-time `check_cache()` was moved behind `interactive()`.
+
+## Fixed
+
+- Missing parenthesis was added in warning text in `check_cache()`.
+
 # rcrisp 0.3.1 - 2025-11-24
 
 ## Changed
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,9 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Missing parenthesis was added in warning text in `check_cache()`.
 - Vignette pre-compilation was updated so that srr tags dropped by `knitr::knit()` are reinserted into the vignettes. Affected vignettes were also recompiled.
+- The input `dem_source` in `get_dem()` was made case insensitive.
 - Fixed typos in test statements.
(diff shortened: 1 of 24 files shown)


R/zzz.R
cforgaci, line 1: @fnattino, I am not entirely sure about this. First, I make sure that there is an option `rcrisp.run_checks` available and set to `FALSE`. Then `check_cache()` is run only in interactive mode or when that option is set to TRUE.

fnattino: Do we need the `.onLoad` part? Could we do `getOption("rcrisp.run_checks", default = FALSE)`, all within the `.onAttach()`?

fnattino: Also I guess that the inclusion of `interactive()` is probably superfluous most of the time (isn't `library(...)`  predominantly done in interactive use?) - but it is probably fine to keep it there for cases where you have the library attached in scripts/vignettes.

R/zzz.R
```diff
@@ -1,4 +1,14 @@
...
+  if (is.null(getOption("rcrisp.run_checks"))) {
+    options(rcrisp.run_checks = FALSE)
+  }
+  invisible()
+}
+
 #' @noRd
 .onAttach <- function(libname, pkgname) {
-  check_cache()
+  if (interactive() || getOption("rcrisp.run_checks")) {
```
fnattino: A very minor point - if the option is only used to check the cache, maybe it makes sense to call it `rcrisp.check_cache` instead?

fnattino (changes requested):
Just a suggestion, let me know if does not make sense!

R/zzz.R
```diff
@@ -1,4 +1,14 @@
+#' @noRd
+.onLoad <- function(libname, pkgname) {
+  if (is.null(getOption("rcrisp.run_checks"))) {
```
cforgaci:
```suggestion
  if (is.null(getOption("rcrisp.check_cache"))) {
```

R/zzz.R
cforgaci:
```suggestion
  if (interactive() || getOption("rcrisp.check_cache")) {
```

R/zzz.R
cforgaci: > Do we need the `.onLoad` part? Could we do `getOption("rcrisp.run_checks", default = FALSE)`, all within the `.onAttach()`?

`.onLoad()` is used the first time a package is called with the namespace operator `::`. `.onAttach()` is not run in that case. Also, [I read](https://forum.posit.co/t/when-to-use-onload-vs-onattach/21953/3) that non-user-set package options are typically in `.onLoad()`. 

On the other hand, your suggestion simplifies this considerably (see the updated code below), so I did just that.

cforgaci: > it is probably fine to keep it there for cases where you have the library attached in scripts/vignettes.

Then I would keep it.

R/zzz.R
```diff
@@ -1,4 +1,14 @@
+#' @noRd
+.onLoad <- function(libname, pkgname) {
+  if (is.null(getOption("rcrisp.check_cache"))) {
+    options(rcrisp.run_checks = FALSE)
```
cforgaci:
```suggestion
    options(rcrisp.check_cache = FALSE)
```

R/zzz.R
```diff
@@ -1,4 +1,14 @@
...
+  }
+  invisible()
+}
+
 #' @noRd
 .onAttach <- function(libname, pkgname) {
-  check_cache()
+  if (interactive() || getOption("rcrisp.check_cache")) {
+    check_cache()
+  }
```
cforgaci:
```suggestion
.onAttach <- function(libname, pkgname) {
  if (interactive() || getOption("rcrisp.check_cache", default = FALSE)) {
    check_cache()
  }
```

cforgaci:
> Just a suggestion, let me know if does not make sense!

@fnattino, thanks for looking into this. I made the changes you suggested.

NEWS.md
```diff
@@ -17,6 +19,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - The OSM retrieval vignette was updated to retrieve CRS with `get_utm_zone()`
 
+## Changed
+
```
cforgaci:
```suggestion
```

cforgaci:
@fnattino I made the changes you requested here. Can you please confirm that this is ready to be merged now?

fnattino (approved):
Yes, @cforgaci looks great to me, thanks!

PR #373, opened 2026-02-26
Increase test coverage

### LINKED ISSUE #372: Increase test coverage in `osmdata.R`
From the rOpenSci review:

> From my perspective the R/osmdata.R could benefit from additional tests.

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Related Issues

- Related Issue #372 
- Closes #372

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes


R/osm.R
```diff
@@ -455,9 +455,13 @@ get_osm_streets <- function(aoi, crs = NULL, highway_values = NULL,
...
-  poly_to_lines <- suppressWarnings(
-    streets$osm_polygons |> sf::st_cast("LINESTRING")
-  )
+  if (!is.null(streets$osm_polygons)) {
+    poly_to_lines <- suppressWarnings(
+      streets$osm_polygons |> sf::st_cast("LINESTRING")
+    )
+  } else {
+    poly_to_lines <- sf::st_as_sf(sf::st_sfc(sf::st_polygon(), crs = crs))
+  }
```
cforgaci, line 552: I am not sure this works, though, as it has not been properly tested.

tests/testthat/test-osmdata.R
```diff
@@ -277,6 +382,82 @@ test_that("If no river surface is found, an empty sfc object is returned", {
...
+    }
+  )
+
+  mock_streets_no_poly <- mock_streets
+  mock_streets_no_poly$osm_polygons <- sf::st_sf(
+    highway = c("primary"),
+    geometry = sf::st_sfc(
+      sf::st_polygon(),
+      crs = crs)
+  )
```
fnattino: If we want to test the case that is now addressed with the fix, shouldn't the polygons actually be NULL?
```suggestion
  mock_streets_no_poly$osm_polygons <- NULL
```

R/osmdata.R
```diff
@@ -455,9 +455,13 @@ get_osm_streets <- function(aoi, crs = NULL, highway_values = NULL,
...
   # Cast polygons (closed streets) into lines
-  poly_to_lines <- suppressWarnings(
-    streets$osm_polygons |> sf::st_cast("LINESTRING")
-  )
+  if (!is.null(streets$osm_polygons)) {
+    poly_to_lines <- suppressWarnings(
+      streets$osm_polygons |> sf::st_cast("LINESTRING")
+    )
+  } else {
+    poly_to_lines <- sf::st_as_sf(sf::st_sfc(sf::st_polygon(), crs = crs))
```
fnattino: I think the geometry type should be linestring here, and OSM data should come in lat/lon:
```suggestion
    poly_to_lines <- sf::st_as_sf(sf::st_sfc(sf::st_linestring(), crs = sf::st_crs("EPSG:4326")))
```

R/osm.R
fnattino: Thanks for catching this! I am actually surprised this has not created problems until now. I guess this field is almost always populated (it's probably enough to have a roundabout) - but having the fall-back plan does not hurt!

fnattino (changes requested):
Thanks a lot for this! Just one small suggestion for (I think) improvement, then it's ready I think!

cforgaci:
> Thanks a lot for this! Just one small suggestion for (I think) improvement, then it's ready I think!

Thanks, @fnattino! A few more CRS fixes, similar to the one you suggested, were needed. Is this okay now?

PR #377, opened 2026-03-05
rOpenSci review base fixes lint issues

cforgaci:
This provides a base with lint fixes before we address the issues raised in https://github.com/ropensci/software-review/issues/718

PR #380, opened 2026-05-19
Restore srr chunks in precompiled vignettes

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Description
As we observed previously in https://github.com/CityRiverSpaces/rcrisp/pull/349 and as pointed out now in https://github.com/ropensci/software-review/issues/718#issuecomment-4488190932, srr chunks were dropped in the precompiled vignettes. As a result, the srr check fails because some srr tags are found only in vignettes. I have updated the precompile script to reinsert the srr chunks dropped by `knittr::knit()` and recompiled the affected vignettes.

## Related Issues

- Related Issue #
- Closes #

## Added/updated tests?

- [ ] Yes
- [x] No, and this is why: this is a documentation update which does not require changes in tests.

## Added entry in changelog?

- [x] Yes

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+# [Unreleased]
+
+## Fixed
+
+- Vignette pre-compilation was updated so that srr tags dropped by `knitr::knit()` are reinserted into the vignettes. Affected vignettes were also recompiled.
+
 # rcrisp 0.3.1 - 2025-11-24
 
 ## Changed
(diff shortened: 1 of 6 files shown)


cforgaci:
@fnattino can you please review this documentation fix?

Note that the failing check should be fixed by #377.

fnattino (approved):
Thanks @cforgaci , I have tested it locally and it works well! I wonder whether one could have also used some nested code cells to have knitr keep the cell on first execution and drop it on subsequent executions (e.g. when building the actual HTML files). But probably not worth to investigate this further, I assume the SRR tags will only be needed until rOpenSci review is over?

cforgaci:
> I wonder whether one could have also used some nested code cells to have knitr keep the cell on first execution and drop it on subsequent executions (e.g. when building the actual HTML files). But probably not worth to investigate this further, I assume the SRR tags will only be needed until rOpenSci review is over?

Thanks, @fnattino, for the quick review. I was also considering something like that. However, I could not find a solution with kintr and this is quite time-sensitive, so I went for the reinsertion approach. I will merge like this if you say it works on your end as well.

PR #409, opened 2026-06-04
Redesign main functions

### LINKED ISSUE #387: MP3 Redesign main function(s) to enable more intuitive workflow
This is an issue raised as part of https://github.com/ropensci/software-review/issues/718

Would this not be addressed by point 2 in https://github.com/CityRiverSpaces/rcrisp/issues/386?

cforgaci:

- [x] Refactor
- [ ] Feature
- [ ] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Description
As suggested in #387, `delineate()` and related functions were refactored to simplify the user interface. As a result, the following functions were created/updated:
- `define_aoi()` is a new function that contains all initial AOI parameters, including the CRS and the various buffers used by `get_osmdata()` and `get_dem()`, as well as the parameter setting the initial corridor method to be used as input in `get_dem()`. Note that I kept the delineation-algorithm-specific parameters `max_iterations`, `capping_method`, and `angle_threshold` in `delineate()`.
- `get_osmdata()` is updated to handle OSM data retrieval in a typical workflow.
- `get_dem()` is updated to handle DEM data retrieval in a typical workflow.
- `delineate()` now has a simpler signature that takes the `aoi` object/list as well as the `osm_data` and `dem` data as required inputs, while allowing the user to choose which delineation to carry out.
- `delineate_city_river()` is a new convenience function that wraps the entire workflow: it takes only a city name and river name as input and returns all delineations by default.

Note that after the solution is confirmed, I will

- take care of srr annotations and function documentation;
- check if any additional tests are needed.

Nevertheless, any suggestions on these are also welcome during the PR review.

## Related Issues

- Related Issue #387 
- Closes #387

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes


R/aoi.R
```diff
@@ -0,0 +1,101 @@
...
+  if (is.null(buildings_buffer)) {
+    buildings_buffer <- 100
+    message(sprintf(
+      paste(
+        "The default `buildings_buffer` of %d m is used",
+        "for riverspace delineation."
+      ),
+      buildings_buffer
+    ))
+  }
```
fnattino: With this approach there is no way to leave these buffer parameters to null (and thus, network and buildings are always retrieved within `get_osmdata`). 

I think it would be more transparent to define these buffer parameters as regular optional input parameters in `define_aoi`, with default values to 3000 and 100 (and drop the messages logged to the output and "hidden" parameter updates). 

However, in order to "deactivate" the retrieval of network or building data we can opt for one of the following options:

1. We add two boolean input parameters to `get_osmdata` (say `network` and `buildings`?) which control whether network and building data are actually retrieved. 

2. We leave it up to the user to set the value of `network_buffer` and/or `buildings_buffer` to `NULL` (or maybe `FALSE`?) here in `define_aoi`.

What do you think?

R/delineate.R
```diff
@@ -33,53 +17,40 @@
...
 #'
-#' # Use custom buildings buffer
-#' delineate(city, river, buildings_buffer = 150, riverspace = TRUE)
+#' # Delineate with defaults
+#' delineate(aoi, osm_data, dem)
 #'
 #' # Provide DEM as input
 #' bucharest_dem <- get_dem_example_data()
-#' delineate(city, river, dem = bucharest_dem)
+#' delineate(aoi, osm_data, dem = bucharest_dem)
```
fnattino: Maybe the following part of the example can be dropped as it does not seem to add anything new with respect to the previous one?
```r
#' # Provide DEM as input
#' bucharest_dem <- get_dem_example_data()
#' delineate(aoi, osm_data, dem = bucharest_dem)
```

R/osm.R
fnattino, line 1: Not top priority, but just a thought: the module is `osmdata.R`, we have the function `get_osmdata` and use the variable `osm_data` (not to clash with the `osmdata` package) - what if we would rename all of these to `osm`?

R/aoi.R
```diff
@@ -0,0 +1,101 @@
...
+
+  list(
+    city_name = city_name,
+    river_name = river_name,
+    bb = bb,
+    crs = crs,
+    corridor_init = corridor_init,
```
fnattino: Not sure about the `corridor_init` parameter - does it belong more here or as input to `delineate`?

R/delineate.R
```diff
@@ -170,3 +109,37 @@ delineate <- function(
 
   delineations
 }
+
+#' Delineate a corridor around a river
```
fnattino, line 137: Very nice way to introduce an easy access point to all the delineations!

R/valley.R
```diff
@@ -54,18 +55,19 @@ default_stac_dem <- list(
...
 #' @srrstats {SP6.1} If specified by the user, the CRS is standardised with
 #'   `as_crs()` before being used to reproject the DEM.
-get_dem <- function(bb, dem_source = "STAC", stac_endpoint = NULL,
-                    stac_collection = NULL, crs = NULL,
-                    force_download = FALSE) {
+get_dem <- function(aoi, osm_data, dem_source = "STAC", stac_endpoint = NULL,
+                    stac_collection = NULL, force_download = FALSE) {
+  # Retrieve dataset on a larger AOI to limit edge effects in downstream
+  # valley delineation
+  aoi_dem <- buffer(osm_data$aoi_network, aoi$dem_buffer)
```
fnattino: The DEM buffer is a buffer around the network AoI (so not around the river centerline as mentioned somewhere in the docstrings). I wonder whether it would make sense to actually define this with respect to the river centerline and modify the default value of this parameter to the current values of `network_buffer` + `dem_buffer` = 3000 m + 2500 m = 5500 m ?

fnattino (commented):
Thanks a lot @cforgaci, I think this looks really great! This reduces significantly the number of input parameters and seems to make everything clearer and better encapsulated. I have left a couple of suggestions.

R/aoi.R
cforgaci: Thanks for pointing this out @fnattino. I removed those implicit parameter updates and made them regular parameters with default values of 3000 and 100, respectively. I also opted for logical input parameters in `get_osmdata`, both set to `TRUE` by default.

R/aoi.R
cforgaci: I moved it back to `delineate()`. It indeed seems to belong more to the delineation algorithm there.

R/osm.R
cforgaci: I renamed all occurrences of `osm_data` and `osmdata` to `osm`

R/valley.R
cforgaci: To me, it makes more sense to keep it as is, since this buffer is meant to increase the DEM around the network, right?

cforgaci:
Thanks @fnattino for the suggestions. I tried to address all. Can you please have another look at this?

R/aoi.R
```diff
@@ -0,0 +1,69 @@
...
+#' @param city_name A character vector of length one
+#' @param river_name A character vector of length one
+#' @param crs The projected Coordinate Reference System (CRS) to use. If not
+#'   provided, the suitable Universal Transverse Mercator (UTM) CRS is selected
+#' @param network_buffer Add a buffer (an integer in meters) around
+#'   river to retrieve additional data (streets, railways, etc.).
+#'   Default is 3000 m.
+#' @param dem_buffer Size of the buffer region (in meters) around the spatial
+#'   network to retrieve the DEM  (only used if `corridor_init` is `"valley"`
+#'   and `dem` is NULL).
```
fnattino:
```suggestion
#'   network to retrieve the DEM.
```

R/aoi.R
```diff
@@ -0,0 +1,69 @@
...
+#' @param buildings_buffer Add a buffer (an integer in meters) around the
+#'   river to retrieve additional data (buildings). Default is 100 m.
+#'
+#' @return A list with delineation parameter values, namely `city_name` and
+#'   `river_name` as character vectors of length one, `bb` as object of class
+#'   [`sf::bbox`][sf::st_bbox], `crs` as object of class [`sf::crs`],
+#'   `corridor_init` as character or numerical vector of length one, and
```
fnattino:
```suggestion
#'   [`sf::bbox`][sf::st_bbox], `crs` as object of class [`sf::crs`], and
```

R/aoi.R
```diff
@@ -0,0 +1,69 @@
...
+#' @examplesIf interactive()
+#' # Get default parameters in AOI
+#' aoi <- define_aoi("Bucharest", "Dâmbovița")
+#'
+#' # Get parameters in AOI with custom CRS
+#' aoi <- define_aoi("Bucharest", "Dâmbovița",
+#'                   crs = "EPSG:3844")  # National projected CRS
+#'
+#' # Get parameters in AOI with non-default initial corridor buffer
+#' aoi <- define_aoi("Bucharest", "Dâmbovița", corridor_init = 1000)
```
fnattino:
```suggestion
```

R/osm.R
fnattino: Should we rename this module as well? (and similarly, `test-osmdata.R`)

R/valley.R
fnattino: OK, let's leave it as with the current meaning (e.g. as a buffer around `aoi_network`)!

R/aoi.R
```diff
@@ -0,0 +1,69 @@
...
+#'
+#' # Get parameters in AOI with non-default buffers
+#' aoi <- define_aoi("Bucharest", "Dâmbovița",
+#'                   network_buffer = 2000,
+#'                   dem_buffer = 2000,
+#'                   buildings_buffer = 150)
+define_aoi <- function(
+  city_name, river_name,
+  crs = NULL,
+  network_buffer = 3000, dem_buffer = 5500, buildings_buffer = 100
```
fnattino: Apologies if I created confusion, but if we keep the meaning of the `dem_buffer` as buffer around the network region (not around the river) as discussed in https://github.com/CityRiverSpaces/rcrisp/pull/409#discussion_r3416762616, this should remain 2500?


```suggestion
  network_buffer = 3000, dem_buffer = 2500, buildings_buffer = 100
```

fnattino (changes requested):
Hi @cforgaci - thanks a lot for considering the feedback. I have noticed a couple of minor issues, feel free to merge afterwards!

R/aoi.R
cforgaci: Correct, thanks. That was left over from a previous attempt I made when trying out your suggestion.

cforgaci:
@fnattino I processed the changes you requested. I also noticed that a few tests were still using the old function signatures, so I updated them as well. Can you please have another look?

fnattino (approved):
Looks great @cforgaci, thanks for all your work on this!

PR #411, opened 2026-06-08
Add distance pre-processing and make standard G2.6 applicable

cforgaci:
## What type of PR is this? (check all applicable)

- [ ] Refactor
- [x] Feature
- [ ] Bug Fix
- [ ] Optimization
- [x] Documentation Update

## Related Issues

- Related Issue #395
- Closes #395

## Added/updated tests?

- [x] Yes
- [ ] No, and this is why: _please replace this line with details on why tests
      have not been included_
- [ ] I need help with writing tests

## Added entry in changelog?
_For user-facing changes, add a line describing the changes in [NEWS.md](/NEWS.md)_

- [x] Yes

PR #412, opened 2026-06-08
Issue messages on CRS conversions

### LINKED ISSUE #396: Reconsider applicability of standard G2.9 about issuing messages for type conversion including metadata
> G2.9 about issuing messages for type conversion including metadata. Coordinate systems allow explicit control throughout, and default to convenient UTM zones, but I think that leaves a lot of scope for issuing conversion messages to help users understand what might be happening. I also think that if [units](https://r-quantities.github.io/units/) are used like I suggested above, then conversions might be even more useful, for example by permitting buffers to be specified in different units (distances and arc degrees, for example).

cforgaci:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description
To be updated after #411 and #409 are merged:
- The `preprocess_distance()` function introduced in #411 also needs to issue a message on type conversion.
- Delineate tests are considerably changed in #409, so the tests in this branch should be checked once again after that branch is merged.

## Related Issues

- Related Issue #396
- Closes #396 

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes


R/utils.R
fnattino, line 1: Here we only temporarily reproject to a projected CRS for buffering (the geometry is projected back to lat/lon if input is geographic). Do you think it is anyway worth to notify user?

fnattino (commented):
Just one comment as I am unsure on one of the two notes. Feel free to merge afterwards!

R/utils.R
cforgaci: As far as I understand, we should not leave any CRS conversion implicit, according to standard G2.9. Another issue here is that we are issuing the same general message multiple times during delineation, which is not very informative. I moved the message to user-facing functions and made them more specific on why the transformation is made. Does this make more sense?

PR #415, opened 2026-06-10
Use cli to show message instead of warning when loading data from cache

### LINKED ISSUE #391: MP7 Remove warnings when loading from cache and tidy screen output
This is an issue raised as part of https://github.com/ropensci/software-review/issues/718

> You shouldn't issue warnings when loading data from cache. These should at most be messages, and they should also be able to be [controlled by users](https://ropensci.org/blog/2024/02/06/verbosity-control-packages/). (Using [the 'cli' package](https://cli.r-lib.org/) is really helpful for controlling output; not a recommendation, but if you want to think about using it, it can really help.)

yehanwu:

- [ ] Refactor
- [ ] Feature
- [ ] Bug Fix
- [x] Optimization
- [ ] Documentation Update

## Description

Replace `warning()` with `cli::cli_inform()` when loading data from cache in `read_data_from_cache()`. 

Add `cli` to Imports in DESCRIPTION

Update tests: `expect_warning` → `expect_message` for cache reads in `test-cache.R`, `test-osmdata.R`, `test-valley.R`

## Related Issues

- Related Issue #391 
- Closes #391

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [ ] Yes


R/cache.R
```diff
@@ -116,7 +116,7 @@ get_rds_filename <- function(...) {
 #' @keywords internal
 read_data_from_cache <- function(filepath, unwrap = FALSE, quiet = FALSE) {
   if (!quiet) {
-    warning(sprintf("Loading data from cache directory: %s", filepath))
+    cli::cli_inform("Loading data from cache directory: {filepath}")
```
cforgaci: cli::cli_alert_info() gives a nicer output. While the issue was raised for the warning in read_data_from_cache(), the same kind of tidy output could be aplied to messages in write_data_to_cache() and clear_cache() as well.

I am not sure if the othe warning in check_cache() should also be replaced with cli::cli_alert_info(). Probably yes. Was there a reason in the early discussion about CRAN that would require us to use a warning here?

R/cache.R
cforgaci:
```suggestion
#' @param quiet Omit message on cache file being loaded
```

cforgaci (changes requested):
Thanks @yehanwu! See a couple of suggestions.

R/cache.R
yehanwu: @fnattino [you mentioned](https://github.com/CityRiverSpaces/rcrisp/pull/235#pullrequestreview-2985881169) that a warning is a better fit in `check_cache()`. Considering the reviewer's comment, would it make sense to make that a message as well?

fnattino: Sounds good to me! Going back to the previous discussion, I thought a warning would make more sense in `check_cache()` to alert the user about potentially old cached data (these could be potentially outdated). But `cli_alert_info()` seems actually a good replacement!

R/cache.R
```diff
@@ -179,19 +179,19 @@ clear_cache <- function(before_date = NULL) {
...
-      message("Cache files successfully removed.")
+      cli::cli_alert_info("Cache files successfully removed.")
     }
   }
 }
 
 #' Check cache
 #'
-#' A warning is raised if the cache size is > 100 MB or if it includes files
+#' A message is raised if the cache size is > 100 MB or if it includes files
```
cforgaci:
```suggestion
#' A message is displayed if the cache size is > 100 MB or if it includes files
```

tests/testthat/test-cache.R
```diff
@@ -133,13 +133,13 @@ test_that("Only cache before given date is removed", {
   expect_true(grepl("tmp2", list.files(cache_dir)[1]))
 })
 
-test_that("Cache checks raise no warnings for small, recent files", {
+test_that("Cache checks raise no messages for small, recent files", {
```
cforgaci:
```suggestion
test_that("Cache checks display no messages for small, recent files", {
```

cforgaci (commented):
Thanks @yehanwu! I made two small suggestions (that messages are "displayed", not "raised"). Please check if there are any other places where this phrasing applies.

cforgaci (approved):
Thanks @yehanwu! I will go ahead and merge this.

Issue #414, opened 2026-06-09
Current delineation for Bucharest different from what is documented in the README and vignettes


cforgaci:
@fnattino, as noticed in https://github.com/CityRiverSpaces/rcrisp/pull/370/changes#r3352247498, delineation is different from what we have documented in the README and in vignettes:

<img width="784" height="318" alt="Image" src="https://github.com/user-attachments/assets/1ecba911-e545-46e9-80ba-9f3576ee344b" />

I have also checked the case of Warsaw, and the differences are considerable:

<img width="422" height="511" alt="Image" src="https://github.com/user-attachments/assets/f5bcc8c8-d5f5-4a22-81d1-909447d10ee4" />

<img width="580" height="610" alt="Image" src="https://github.com/user-attachments/assets/34f10ddb-5846-4ef1-8ae8-af23e4042517" />

Any thoughts on what change in the package led to this? My guesses are: changes in network pre-processing or some change in corridor (or maybe valley) delineation. Let me know if you have any hints on how to diagnose this. It might be that the delineation remains as it is now, but I would like to understand what led to the change.

cforgaci:
One thing that might help diagnose problems like this is a quick way to plot the river, the delineated valley, the network and the delineation. The `plot.delineation` method we are planning to implement should do that.

cforgaci:
And beyond visual inspection (user-oriented), we should test this to avoid changes in the codebase that lead to considerable changes in the resulting delineation. I suggest:

- A delineation regression test, that is, checking against a known case within a certain tolerance.
- Property-based tests to catch obviously wrong output (we might have some of these already)

fnattino:
Hi @cforgaci, this is indeed weird. What I suspect, however, is that the (main?) source of differences is not changes in the code, but changes in the input data. To check this, I went back to the commit at which the previous version of the figure was generated and rerun the code snippet in the README file - this produced the new figure you obtained (right panel above). On the contrary, using the example OSM data that we uploaded to Zenodo together with the code that is currently in the main branch produces a corridor that is identical to the older delineation (left panel above). 

So maybe comparing the old input data from Zenodo vs the new input data recently downloaded from OSM can shed some light on why the delineation changed so drastically?

fnattino:
> I suggest:
> 
>   * A delineation regression test, that is, checking against a known case within a certain tolerance.

[...]

PR #419, opened 2026-06-22
Retrieve OSM river by id

cforgaci:

- [ ] Refactor
- [ ] Feature
- [x] Bug Fix
- [ ] Optimization
- [ ] Documentation Update

## Related Issues

- Related Issue #414 
- Closes #

## Added/updated tests?

- [x] Yes
      have not been included_

## Added entry in changelog?

- [x] Yes


man/check_cache.Rd
cforgaci, line 1: We forgot to document these in #415, so it is off-topic here. Can we leave it in this PR?

man/read_data_from_cache.Rd
cforgaci, line 1: We forgot to document these in #415, so it is off-topic here. Can we leave it in this PR?

R/osm.R
```diff
@@ -160,9 +228,13 @@ get_osm <- function(aoi,
...
-  # Retrieve the river center line and surface
+  # Retrieve the river center line and surface, expanding the crop area by the
+  # combined network + DEM buffer so the centreline covers the full analysis
+  # extent used by downstream steps
   river_centerline <- get_osm_river_centerline(
-    bb, aoi$river_name, crs = crs, force_download = force_download
+    bb, aoi$river_name, crs = crs,
+    buffer_distance = aoi$network_buffer + aoi$dem_buffer,
+    force_download = force_download
   )
```
cforgaci, line 238: I think we have encountered the issue of cropping the OSM river centerline, but I do not remember why we did not do it after all. I am adding a(nother...) buffer parameter here to extend the length of the cropped river by the network + dem buffers. No river centerline extension leads to slightly different valley edges and thus delineations in the end segments of the corridor. The buffer overcomes that edge effect. Now, I am not sure if a new buffer parameter is really useful.

R/osm.R
```diff
@@ -301,35 +377,51 @@ get_osm_city_boundary <- function(bb, city_name, crs = NULL, multiple = FALSE,
...
+
+  # Fix invalid geometries, crop to crop_area (clips straddling geometries
+  # rather than keeping them whole), and union
+  river_centerline <- suppressWarnings(
+    river_centerline_lines |>
+      check_invalid_geometry() |>
+      sf::st_crop(crop_area)
+  ) |>
     sf::st_geometry() |>
     sf::st_union()
```
cforgaci, line 424: This implements the approach from the shiny app

cforgaci:
@fnattino I tried to implement the approach from the shiny app, as you suggested. Let me know what you think.

tests/testthat/test-osm.R
cforgaci:
```suggestion
# All functions using [`osmdata_as_sf()`] or [`osmdata_as_sf_by_id()`] store and
# data in a cache folder and read data from there when already available. In
# order not to mess up with the user cache directory, we set up a temporary cache
# folder that already contains some datasets, only used for testing purposes.
# This is achieved via the [`temp_cache_dir()`] helper function, which should be
# called in each test that does not mock [`osmdata_as_sf()`] or
# [`osmdata_as_sf_by_id()`].
```

Issue #399, opened 2026-05-28
Reconsider applicabilty of standard G5.7 on algorithm performance tests


cforgaci:
> G5.7 if the authors decide to add performance/scale expectations and corresponding tests (even if only as extended tests).

> G5.7 As mentioned in https://github.com/ropensci/software-review/issues/718#issuecomment-3928093410, some kind of demonstration of scaling would really help. The algorithm sometimes seems quite slow, and I suspect scaling with data size may be strongly non-linear. It would really help people to know what to expect. And once you've gone through some process of documenting that, putting in a few concrete tests should be straightforward.

PR #421, opened 2026-06-29
Implement and document benchmarking

### LINKED ISSUE #369: Document algorithmic scaling/performance boundaries
From rOpenSci review:

> I did not see much explicit discussion of algorithmic scaling/performance boundaries. That may be acceptable, but for a Gold target it would be helpful to at least document expected “city size” / data-size constraints, and where runtime is expected to be dominated (OSM download, DEM retrieval, cost-distance, etc.).

### LINKED ISSUE #394: Reconsider applicability of standards G1.5 / G1.6 on performance claims
> G1.5 / G1.6 if future versions begin making explicit performance claims or if alternative implementations become available for benchmarking.

fnattino:
I think there are different levels at which we can try to #369  and #399. At the very simplest leve, we could only state that the OSM data retrieval is by far the most time-consuming part, and that for cities like Paris and London we could not run riverspace delineation, with the bottleneck being the building retrival.

If we want (and can afford) to spend more time on this, we could try to profile the typical delineation workflow on a few cities. We could measure the time of execution for the various steps (e.g. OSM data retrieval, DEM data retrieval, corridor delineation, segmentation, etc.). 

Even for a single city, this exercise would show the difference in timing for the different steps. By comparing the timings of a given step for different cities we could in addition get a feeling for the scaling of the application.

I am adding here a script which I have quickly generated with the help of Claude (I have mentioned it in the header for transparency). The script loads the city-river pairs from `benchmark/data/city_rivers.csv`, run the delineations timing for all the relevant steps, and write out the partial and totala timings in a CSV file. In implements some retry logic, forces download of data to avoid use of cached data, and, for failed delineations, it should simply report “success = FALSE”.

We could run this or a similar script on a few cities. I have added here an input files with all the city rivers we wanted to consider for the dataset. It is probably not necessary to run this for all of them, but we could try a selection (say between 5 and 10 ciy rivers?) considering different sizes? 

We could use the produced data to make two plots: 

1) plot partial timings vs cities as a stacked bar chart, which should show that the OSM retrieval is by far the most expensive step. For reference, for Bucharest/Dambovita (which is actually not in the list):
  ```csv
  "city_name","river_name","step","elapsed_s","status","error"
  "Bucharest","Dâmbovița","get_osm",46.185,"success",""
  "Bucharest","Dâmbovița","get_dem",4.696,"success",""
  "Bucharest","Dâmbovița","delineate_valley",0.818,"success",""
  "Bucharest","Dâmbovița","as_network",3.414,"success",""
  "Bucharest","Dâmbovița","delineate_corridor",1.476,"success",""
  "Bucharest","Dâmbovița","delineate_segments",3.455,"success",""
  "Bucharest","Dâmbovița","delineate_riverspace",2.452,"success",""
  "Bucharest","Dâmbovița","total",62.718,"success",""
  ```

2) Plot total timings vs some quantity that is a proxy of city/river size (which could be the corridor area?) as a scatter plot, maybe adding the name of the cities next to the dots in the plot? This would give some of how long the application can take. We could actually also make the same plot also for all the partial timings, which would additionally give some ideas of where the bottleneck/worst scaling behaviour is.

Not sure where these plots could be included - it could maybe be a vignette? 

What do you think? Would you have time to run this script or something similar on a few cities and try to produce the visualizations? Hope this is useful..

We could also think about adding a short version of the script transformed in a performance test? I would do this only for local execution, so with `skip_on_cran()` and `skip_on_ci()`. I would also skip the download part, and maybe run on Bucharest/Dambovita using the example osm and dem data. For that, we could use a package like `bench` to do run timings several times. But this I think would be extra work, one reviewer suggested to add this, we could mention this could be one of our next steps..

A caveat: when trying to run the script myself, I encountered some issues with the OSM server, with calls timing out continuously. I am not sure whether this is some sort of issue of the default overpass API server. Adding the following lines to the script made it work smoothly (I got this alternative URL from [here](https://wiki.openstreetmap.org/wiki/Overpass_API#Public_Overpass_API_instances)): 
```r
library(osmdata)
url <- "https://maps.mail.ru/osm/tools/overpass/api/interpreter"
set_overpass_url(url)
```


fnattino:
Feel free to take over this branch/PR!

cforgaci:
@fnattino many thanks for the guidance and for kicking off the work on this! Following your suggestions, I plan to document algorithmic scaling in a vignette and to create a performance test.

